PR #5, opened 2022-11-11
Allow extrapolation of step time series for the logical check

evertoncolling:
This PR includes an optional `endTime` parameter to the **Logical Check**. This allows the routine to extrapolate a step time series up to the `endTime` before executing the logical check operation. If a user provides a value for `endTime`, but the provided time series is not of type `step`, we don't extrapolate until the `endTime`.

### DIFF
--- a/Cognite.DataProcessing/DataSampling.cs
+++ b/Cognite.DataProcessing/DataSampling.cs
@@ -98,7 +98,7 @@ namespace Cognite.DataProcessing
         /// <param name="ts">The time series to evaluate</param>
         /// <param name="threshold">The threshold to use for the logical check</param>
         /// <param name="check">The logical check to use</param>
-        /// <param name="endTime">The logical check to use</param>
+        /// <param name="endTime">End time in milliseconds since Epoch</param>
         /// <returns>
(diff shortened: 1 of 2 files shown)


Cognite.DataProcessing/DataSampling.cs
```diff
@@ -98,16 +98,22 @@ public static TimeSeriesData UnionLogicTimeSeries(TimeSeriesData ts1, TimeSeries
         /// <param name="ts">The time series to evaluate</param>
         /// <param name="threshold">The threshold to use for the logical check</param>
         /// <param name="check">The logical check to use</param>
+        /// <param name="endTime">The logical check to use</param>
```
lprovensi, line 101:
```suggestion
        /// <param name="endTime">The logical check to use, in milliseconds since Epoch</param>
```

lprovensi (approved):
Nice :)

Cognite.DataProcessing/DataSampling.cs
evertoncolling: Very good catch!

PR #10, opened 2023-01-02
Base class for implementing connectors

lprovensi:
Abstract class with heartbeat reporting functionality


evertoncolling (approved):
LGTM 👍

PR #11, opened 2023-01-05
Base client class for connecting to COM automation servers

lprovensi:
Some simulators such as PROSPER and PetroSIM can be accessed through COM automation servers. This is a base class to implement connectors as automation clients.
At the moment, there is not a good solution to test this in CI/CD, thus there are no tests in this PR


evertoncolling (approved):
Great to have the COM interface initialisation captured in a generic manner! 👍

PR #42, opened 2023-05-24
Enable simulation runs API

polomani:
Users can enable the use of the simulator runs API by setting in the config.yaml:
```
connector:
     use-simulators-api: true
```

- test is disabled until we have 
1. persistence for the eventId on the simulation run object
2. runTime parameter that allows to choose time range of the simulation
3. a bit more filters (or sort) on the simulation runs list

### DIFF
--- a/Cognite.Simulator.Tests/UtilsTests/SimulationRunnerTest.cs
+++ b/Cognite.Simulator.Tests/UtilsTests/SimulationRunnerTest.cs
@@ -3,7 +3,6 @@ using Cognite.Extractor.Utils;
 using Cognite.Simulator.Extensions;
 using Cognite.Simulator.Utils;
 using CogniteSdk;
-using CogniteSdk.Alpha;
 using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.Logging;
 using System;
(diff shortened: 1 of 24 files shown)


Cognite.Simulator.Tests/UtilsTests/SimulationRunnerTest.cs
```diff
@@ -231,6 +250,160 @@ public async Task TestSimulationRunnerBase()
...
+                // Run the simulation runner and verify that the event above was picked up for execution
+                using var linkedTokenSource2 = CancellationTokenSource.CreateLinkedTokenSource(source.Token);
+                var linkedToken2 = linkedTokenSource2.Token;
+                linkedTokenSource2.CancelAfter(TimeSpan.FromSeconds(15));
+                var taskList2 = new List<Task> { runner.Run(linkedToken2) };
+                await taskList2.RunAll(linkedTokenSource2).ConfigureAwait(false);
+
+                Assert.True(runner.MetadataInitialized);
+
+                // Uncomment when we have runTime param in place
```
polomani, line 337: not runTime but validatioEndTime

polomani: change aggregation type for stepInterpolation and in that case the test will succeed regardless of wether we have dps or not

Cognite.Simulator.Utils/SimulationRunnerBase.cs
```diff
@@ -411,68 +581,98 @@ public async Task Run(CancellationToken token)
...
+            if (simEv.HasSimulationRun)
+            {
+                if (!simEv.Run.EventId.HasValue)
+                {
+                    _logger.LogDebug("Simulation run has no Event associated with it {Id}", simEv.Run.Id);
+                    return;
+                }
+                simEvent = await _cdfEvents.GetAsync(simEv.Run.EventId.Value, token);
             }
+            else
```
polomani, line 671: fix warning (for example docs on public methods)

lprovensi (approved):
Approved, as this was discussed during office hours

PR #48, opened 2023-06-13
Connector can register multiple simulators with CDF

lprovensi:
Previously, the connector could register multiple simulators, but it was not able to update the information individually. That is, all simulators would have the same 'simulator version'.
Also, a new field was added to indicate if the connector supports the SimInt API. This is disabled by default.


polomani (approved):
LGTM

PR #49, opened 2023-06-13
[DEGR-2510] Allow commands without type

lprovensi:
OpenServer commands, for instance, do not need a type. They are executed with an 'address' as argument


polomani (approved):
LGTM

sagar-thalwar (approved):

PR #56, opened 2023-07-03
Connector check for license every 1 hour [DEGR-2670]

kbrattli:
**Related**
This breaks UI but this fixes it: https://github.com/cognitedata/application-services/pull/5371
Display in frontend: https://github.com/cognitedata/applications/pull/3870

- [DWSIM](https://github.com/cognitedata/dwsim-connector-dotnet/pull/20) (TODO: Delete this branch)
- [Symmetry](https://github.com/cognitedata/symmetry-connector-dotnet/pull/22)
----------------------------------------------------
**Summary**
Adds a Timestamp to the Sequence of the last time there was a confirmed available license for the simulator in the connector. Has to be enabled in the Config.yaml like this:
```
connector:
  name-prefix: "symmetry"
  add-machine-name-suffix: false
  license-check:
    enabled: true
    interval: 3600
```
- `LicenseCheck()` in **ConnectorBase.cs**: Calls the `UpdateIntegrationRows()` once every hour with the bool parameter licenseCheck set to True. This only happens if the abstract method`CheckLicensesStatus()` is true, this method will be overwritten by each simulator.
- `UpdateIntegrationRows():` When licenseCheck parameter is True, call `UpdateSimulatorIntegrationsLicenseTimestamp()`
- `UpdateSimulatorIntegrationsLicenseTimestamp()` in **SequencesExtensions.cs**: Updates the sequences row related to license timestamp to now.
----------------------
**How to add license check to each connector**
- The method `CheckLicenseStatus()` from `ConnectorBase` must be overriden in each connector.
- Important to add it in both `Init()` and `Run()` like this:
```
public override async Task Init(CancellationToken token)
    {
        _symmetry.TestConnection();
        //TODO: Check the connection with the simulator
        var extraInfo = new Dictionary<string, string>();
        extraInfo[SimulatorIntegrationSequenceRows.SimulatorVersion] =
            "N/A"; //TODO: Find the simulator version 

        await EnsureSimulatorIntegrationsSequencesExists(token);
        await UpdateIntegrationRows(true, token); //--------------- Pay attention to this line
        await _modelLibrary.Init(token);
        await _configurationLibrary.Init(token);
        
    }
    public override async Task Run(CancellationToken token)
    {
        _logger.LogDebug("Connector started, sending status to CDF every {Interval} seconds",
            _config.Connector.StatusInterval);
        try
        {
            //_symmetry.ClearCallbackLogMessages();
            var linkedTokenSource = CancellationTokenSource.CreateLinkedTokenSource(token);
            var linkedToken = linkedTokenSource.Token;
            var modelLibTasks = _modelLibrary.GetRunTasks(token);
            var configLibTasks = _configurationLibrary.GetRunTasks(token);
            var taskList = new List<Task> { Heartbeat(linkedToken) };
            taskList.AddRange(modelLibTasks);
            taskList.AddRange(configLibTasks);
            taskList.Add(_simulationScheduler.Run(linkedToken));
            taskList.Add(_simulationRunner.Run(linkedToken));
            if (ShouldLicenseCheck())   //----------------------------- Pay attention to this line
            {
                taskList.Add(LicenseCheck(linkedToken));
            }
            taskList.Add(_pipeline.PipelineUpdate(linkedToken));
            await taskList.RunAll(linkedTokenSource);
        }
        catch (OperationCanceledException)
        {
            _logger.LogWarning("Cancellation requested. Exiting the connector");
        }


    }
```

**License Check is added in these connectors**
---------------------------
- [ ] Symmetry
- [ ] SimConnect
- [ ] PetroSIM
- [x] DWSIM (Does not have license)


Cognite.Simulator.Extensions/SequencesExtensions.cs
```diff
@@ -157,20 +157,29 @@ public static class SequencesExtensions
...
             var rowsToCreate = new List<SequenceDataCreate>();
-            var rowData = new Dictionary<string, string>
-                {
-                    { SimulatorIntegrationSequenceRows.Heartbeat, $"{DateTime.UtcNow.ToUnixTimeMilliseconds()}" }
-                };
+            var rowData = new Dictionary<string, string>();
+
+            if (updateLicense is true)
+            {
+                rowData.Add(SimulatorIntegrationSequenceRows.LicenseTimestamp, $"{DateTime.UtcNow.ToUnixTimeMilliseconds()}");
```
kbrattli, line 176: TODO: Convert to separate functions. Also consider renaming to make it more clear

kbrattli: Ensure row is going to right column

Cognite.Simulator.Utils/Configuration.cs
```diff
@@ -84,6 +84,12 @@ public class ConnectorConfig
         /// </summary>
         public int StatusInterval { get; set; } = 10;
 
+        /// <summary>
+        /// The connector will check for license updates with this interval (in seconds)
+        /// </summary>
+        public int LicenseUpdateInterval { get; set; } = 30;
```
sagar-thalwar, line 90: Suggestion: By default, we seem to be checking for license for every 30 seconds, I feel this needs to be purely driven by config. If the config does not contain `license-update-interval`, then we should not be checking for license. Also we can keep minimum value as at least one hour?
@evertoncolling your thoughts?

kbrattli: Changed LicenseUpdateInterval to 1 hour and added logic in GetLicenseCheckInterval() so that it would be minimum 1 hour. Also added EnableLicenseCheck in config so that check for license can be optional after it is implemented in each connector.

evertoncolling: I agree with the above. We should have a flag on the config file specifying whether to perform the check or not (`license-check` -> `enabled`), if this `enabled` flag is not present we default to not check the license. Additionally we should have an optional frequency parameter in the config file that specifies how frequent we check for a license (`license-check` -> `frequency`). If this `frequency` parameter is not present, we default to checking once per hour.

As a summary we should have 2 new config options (suggested names as below):
```yaml
license-check:
 enabled: true
 frequency: 60
```

Cognite.Simulator.Utils/ConnectorBase.cs
```diff
@@ -32,6 +33,8 @@ public abstract class ConnectorBase
         private readonly ILogger<ConnectorBase> _logger;
         private readonly ConnectorConfig _config;
 
+        private string LastLicenseCheckTimestamp { get; set; } = "0";
```
polomani, line 36: this should be long, convert it to string upon writing to the sequence

polomani:
@kbrattli is this ready to be reviewed? I see some todos still

kbrattli:
@polomani Yes. Only thing i would change is that Interval now is in seconds instead of minutes. But this makes it easier to test the connectors. I use VM to test locally, i can give you demo if you want.

One maybe bad part with this design is that if you restart the connector, it will wipe the previous licenseStatus and licenseTimestamp and then wait one hour before you get a new one.

Cognite.Simulator.Utils/ConnectorBase.cs
kbrattli: If it should end as a string anyway, it seems easier for me to just make it as a string from the beginning

polomani:
@kbrattli standard unit for time is a second, so it should be better to keep it, but also to limit the minimal value there (which is what you already doing)

kbrattli:
@polomani The problem is that when you restart it will run init() all over again, and there is no way for the program to know if it has been run in the past, so the init() will assume there is no previous sequence and overwrite everything.

So a solution could be that you just check for license and update it in the init() aswell. I have not done it, but its probably possible. But this has to be implemented for each connector separately. It's just that my "template" solution for each connector will have the issue i mentioned (in the text: "How to add License check for each connector").

Cognite.Simulator.Utils/ConnectorBase.cs
polomani: > If it should end as a string anyway
This is just because the persistence structure we use only allows string at the moment.
Normally, the timestamp is in fact long, so we should keep it that way up until it is serialized.
This allows us to swap to a different persistence layer (which allows long values) in the future easily.

Overall it's better to not shoehorn the data based on serialisation/deserialisation logic if possible.
(otherwise everything would have been JSON for us) :)

Cognite.Simulator.Utils/ConnectorBase.cs
```diff
@@ -228,6 +266,30 @@ await UpdateIntegrationRows(false, token)
...
+        /// <summary>
+        /// This method should be overridden in each specific Connector class.
+        /// </summary>
+        /// <returns>True if the simulator has a valid license, false otherwise.</returns>
+        public virtual bool CheckLicenseStatus() // make this abstract?
+        {
+            return true;
+        }
+        /// <summary>
+        ///
```
polomani, line 278: maybe add some description?

Cognite.Simulator.Utils/ConnectorBase.cs
```diff
@@ -177,8 +199,22 @@ protected async Task EnsureSimulatorIntegrationsSequencesExists(CancellationToke
...
         {
+            if (licenseCheck is true)
+            {
+                if (init is false)
+                {
+                    LastLicenseCheckTimestamp = DateTime.UtcNow.ToUnixTimeMilliseconds();
+                    LastLicenseCheckResult = licenseCheckStatus;
+                }
+                else {
+                    LastLicenseCheckResult = "Not checked yet.";
```
polomani, line 214:
```suggestion
                    LastLicenseCheckResult = "Not checked yet";
```

Cognite.Simulator.Extensions/Types.cs
```diff
@@ -418,6 +418,11 @@ public class SimulatorIntegrationUpdate : SimulatorIntegration
         /// </summary>
         public bool SimulatorApiEnabled { get; set; }
         
+        /// <summary>
+        /// Flag indicating if there
```
polomani, line 422: unfinished senstence

Cognite.Simulator.Utils/ConnectorBase.cs
```diff
@@ -177,8 +199,22 @@ protected async Task EnsureSimulatorIntegrationsSequencesExists(CancellationToke
         /// </summary>
         protected async Task UpdateIntegrationRows(
             bool init,
-            CancellationToken token)
+            CancellationToken token,
+            bool licenseCheck = false,
+            string licenseCheckStatus = "")
         {
+            if (licenseCheck is true)
```
polomani, line 206: assigning values inside this if statement should be moved into `public async Task LicenseCheck(CancellationToken token)`.
init value for the status could be "Not checked yet", after one hour it'll be overwritten.

this makes this method more lightweight, only reading the current LastLicenseCheckTimestamp/LastLicenseCheckResult values instead.

polomani (commented):
Looks good, I added some comments around some smaller improvements

Cognite.Simulator.Extensions/SequencesExtensions.cs
```diff
@@ -157,27 +157,28 @@ public static class SequencesExtensions
...
+
+            rowData.Add(SimulatorIntegrationSequenceRows.Heartbeat, $"{DateTime.UtcNow.ToUnixTimeMilliseconds()}");
+            rowData.Add(SimulatorIntegrationSequenceRows.LicenseTimestamp, $"{lastLicenseCheckTimestamp}");
+            rowData.Add(SimulatorIntegrationSequenceRows.LicenseStatus, lastLicenseCheckResult);
+
             if (init)
             {
-                if (update == null)
-                {
-                    throw new ArgumentNullException(nameof(update));
```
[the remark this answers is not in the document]
kbrattli: @polomani It is impossible for init to be false and update to be null because of how they are made in a previous function.
`var update = init ?
                    new SimulatorIntegrationUpdate
                    {
                        Simulator = simulator.Name,
                        DataSetId = simulator.DataSetId,
                        ConnectorName = GetConnectorName(),
                        ConnectorVersion = GetConnectorVersion(),
                        SimulatorVersion = GetSimulatorVersion(simulator.Name),
                        ExtraInformation = GetExtraInformation(simulator.Name),
                        SimulatorApiEnabled = ApiEnabled()
                    }
                    : null;
                    await sequences.UpdateSimulatorIntegrationsHeartbeat(
                        _simulatorSequenceIds[simulator.Name],
                        init,
                        update,
                        token).ConfigureAwait(false);
                }`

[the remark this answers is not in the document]
polomani: @kbrattli Yes, but we get a warning, it's better to get rid of it
Screenshot 2023-07-24 at 14 07 49

Cognite.Simulator.Utils/ConnectorBase.cs
kbrattli: Fixed but had to keep this in the function: 
`if (licenseCheck && init)
            {
                LastLicenseCheckResult = "Not checked yet";
            }`

Cognite.Simulator.Utils/ConnectorBase.cs
```diff
@@ -32,6 +33,9 @@ public abstract class ConnectorBase
         private readonly ILogger<ConnectorBase> _logger;
         private readonly ConnectorConfig _config;
 
+        private long LastLicenseCheckTimestamp { get; set; } = 0;
```
polomani, line 36: noticed now, you need to remove the `= 0;` (warning below)

Cognite.Simulator.Utils/ConnectorBase.cs
```diff
@@ -177,7 +200,8 @@ protected async Task EnsureSimulatorIntegrationsSequencesExists(CancellationToke
         /// </summary>
         protected async Task UpdateIntegrationRows(
             bool init,
-            CancellationToken token)
+            CancellationToken token,
+            bool licenseCheck = false)
```
polomani, line 204: this isn't used

Cognite.Simulator.Extensions/SequencesExtensions.cs
```diff
@@ -157,27 +157,32 @@ public static class SequencesExtensions
         /// <param name="sequenceExternalId">Simulator integration sequence external id</param>
         /// <param name="update">Data to be updated, if init is set to true</param>
         /// <param name="token">Cancellation token</param>
+        /// <param name="lastLicenseCheckTimestamp"></param>
+        /// <param name="lastLicenseCheckResult"></param>
```
polomani, line 161: lacking description

Cognite.Simulator.Utils/ConnectorBase.cs
```diff
@@ -51,6 +55,7 @@ public abstract class ConnectorBase
             _simulatorSequenceIds = new Dictionary<string, string>();
             _logger = logger;
             _config = config;
+            LastLicenseCheckResult = ShouldLicenseCheck() ? "Not checked yet" : "License check disabled";
```
polomani, line 58: ShouldLicenseCheck is virtual and prob won't be called from here, maybe make it just "protected"?

Cognite.Simulator.Utils/Configuration.cs
```diff
@@ -177,4 +182,20 @@ public class PipelineNotificationConfig
...
+    {
+        /// <summary>
+        /// The connector will check for license updates with this interval (in seconds)
+        /// </summary>
+        public int Interval { get; set; } = 3600;
+
+        /// <summary>
+        /// Only check for license if this is set to true
+        /// </summary>
+        public bool Enabled { get; set; } = false;
```
polomani, line 199:
```suggestion
        public bool Enabled { get; set; };
```

Cognite.Simulator.Utils/Configuration.cs
```diff
@@ -177,4 +182,20 @@ public class PipelineNotificationConfig
...
+    {
+        /// <summary>
+        /// The connector will check for license updates with this interval (in seconds)
+        /// </summary>
+        public int Interval { get; set; } = 3600;
+
+        /// <summary>
+        /// Only check for license if this is set to true
+        /// </summary>
+        public bool Enabled { get; set; };
```
polomani, line 199:
```suggestion
        public bool Enabled { get; set; }
```

PR #59, opened 2023-07-07
Add support for other OSs

abdullah-cognite:
This is required to get the DWSIM connector to work inside a unix docker container


Cognite.Simulator.Utils/Automation/AutomationUtils.cs
```diff
@@ -51,6 +51,10 @@ public void Initialize()
             if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
             {
                 Server = ActivateAutomationServer();
+            } else if (RuntimeInformation.IsOSPlatform(OSPlatform.Linux) || RuntimeInformation.IsOSPlatform(OSPlatform.OSX))
```
polomani, line 54: this enables it for all of them, but maybe this could be a property on DwsimAutomationConfig instead? so we can enable it only for Dwsim for example

PR #62, opened 2023-08-07
Add support for simulator connector status

abdullah-cognite:
Adds 2 new fields to the simulator integration sequence. 1: ConnectorStatus and 2: ConnectorStatusTimestamp

The ConnectorStatus field would be in either of the 2 states:
- "RUNNING_CALCULATION" when a calculation is running. 
- "IDLE" when a calculation is not running. 

Screenshot 2023-08-07 at 10 48 22


Cognite.Simulator.Extensions/SequencesExtensions.cs
```diff
@@ -566,6 +567,104 @@ public static string[] GetStringValues(this SequenceRow row)
             return result.ToArray();
         }
 
+        /// <summary>
+        // Simply upserts a key value pair into a key value pair based sequence
```
polomani, line 571:
```suggestion
        /// Simply upserts a key value pair into a key value pair based sequence
```

Cognite.Simulator.Utils/SimulationRunnerBase.cs
```diff
@@ -396,6 +405,25 @@ public async Task Run(CancellationToken token)
...
+            
+            try
+            {
+                if (sequenceExternalId == "" && _simulators.Count > 0) {
+                    SimulatorConfig item = _simulators[0]; // Retrieve the first item
+                    sequenceExternalId = await SequencesExtensions.GetSequenceExternalId(sequences, item.Name, item.DataSetId, _connectorConfig.GetConnectorName(), token);
+                }
+
+                await SequencesExtensions.UpsertItemInKVPSequence(_cdfSequences, sequenceExternalId, SimulatorIntegrationSequenceRows.ConnectorStatus, runStatus, token);
+            }
```
polomani, line 419: check the warnings generated by github.

Cognite.Simulator.Utils/SimulationRunnerBase.cs
```diff
@@ -276,6 +282,9 @@ public async Task Run(CancellationToken token)
                             startTime, 
                             e, 
                             token).ConfigureAwait(false);
+                        
+                        PublishSimulationRunStatus("", token);
```
polomani, line 286: as a an alternative we could keep this status in memory and write it back on the heartbeat.
Pros:
1. less code to maintain (no need of UpsertItemInKVPSequence)
Cons:
1. not that granular, but we are only interested in the long-running calculations anyway

Wdyt?

abdullah-cognite: Yep, I thought about this as well. The problem I was facing was creating an in memory resource, that both the `ConnectorBase` and `SimulationRunnerBase` could access.

polomani: @abdullah-cognite what's the problem exactly?

Cognite.Simulator.Utils/SimulationRunnerBase.cs
```diff
@@ -53,6 +53,9 @@ public abstract class SimulationRunnerBase<T, U, V>
         /// </summary>
         protected IConfigurationProvider<U, V> ConfigurationLibrary { get; }
 
+        private string sequenceExternalId;
```
polomani, line 56:
```suggestion
        private string? sequenceExternalId;
```
we can make this prop nullable, then no need of assigning / comparing with the empty string.

polomani: @lprovensi do you know if there are any consequences for enabling nullable types?

Cognite.Simulator.Extensions/DataModel.cs
```diff
@@ -355,6 +355,17 @@ public static class SimulatorIntegrationSequenceRows
...
+
+        /// <summary>
+        /// Status of the connector (indicates whether a calculation is running or not)
+        /// </summary>
+        public const string ConnectorStatus = "connectorStatus";
+
+        /// <summary>
+        /// Timestamp of when the connector went into the new connectorStatus state
+        /// </summary>        
+        public const string ConnectorStatusTimestamp = "connectorStatusTimestamp";
```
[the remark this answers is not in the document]
abdullah-cognite: done

polomani (approved):
LGTM, the complexity should go away once we move to the new API

Cognite.Simulator.Utils/SimulationRunnerBase.cs
lprovensi: @polomani @abdullah-cognite I'm not sure. You would have to enable nullable `<Nullable>enable</Nullable>` and also I think bump the language version `<LangVersion>8.0</LangVersion>`, because .net standard 2.0 does not support it. You can try to compile and run with those to see if it works

Cognite.Simulator.Extensions/SequencesExtensions.cs
```diff
@@ -566,6 +567,106 @@ public static string[] GetStringValues(this SequenceRow row)
...
+        /// Gets the externalId of a SimulatorIntegrations sequence (This function can throw an exception)
+        /// </summary>
+        /// <param name="sequences">CDF Sequence resource</param>
+        /// <param name="simulatorName">The simulator name.</param>
+        /// <param name="simulatorDataSetId">The simulator datasetId</param>
+        /// <param name="connectorName">The connector Name</param>
+        /// <param name="token">The cancellation token</param>
+        /// <returns>externalId string</returns>
+        public static async Task<string> GetSequenceExternalId(this SequencesResource sequences, string simulatorName, long simulatorDataSetId, string connectorName, CancellationToken token) {
+             var _simulatorSequenceIds = new Dictionary<string, string>();
```
lprovensi, line 649: Indentation here, also use _ to prefix only private class members (Not a strict requirement, but more to conform with the conventions we already started with)

Cognite.Simulator.Extensions/SequencesExtensions.cs
```diff
@@ -566,6 +567,106 @@ public static string[] GetStringValues(this SequenceRow row)
...
+            List<string> _simulators = new List<string> { simulatorName };
+            var simulatorsDict = _simulators.Select(
```
lprovensi, line 651: Since it is only one item (`simulationName`), this `Select` is not necessary, and the `_simulators` list is also unnecessary

lprovensi (approved):
A few code improvements suggested but, I looks like a good solution! 👍

PR #78, opened 2023-11-14
feat: enhance cleanup of temporary files [POSFP-73]

abdullah-cognite:
This PR now creates a separate directory for each `model`, so if you're going to generate VLP or GAPRS files for a model in GAP or PROSPER, they will be stored under that folder. Later on, if the model is deleted or updated from the UI, the entire folder will be deleted, so each Connector wont have to take care of deletion of these files. Previously, we were not tracking these files anywhere, so they were just lying around even after the model was deleted or updated.

Jira : https://cognitedata.atlassian.net/browse/POFSP-73


Cognite.Simulator.Utils/FileLibrary.cs
```diff
@@ -215,6 +215,14 @@ public IEnumerable<Task> GetRunTasks(CancellationToken token)
             return new List<Task> { SaveStates(token), SearchAndDownloadFiles(token) };
         }
 
+        public void CreateDirectoryIfNotExists(string directoryPath)
```
lprovensi, line 218: Consider making it private, if only using internal. Else add XML documentation

Cognite.Simulator.Utils/FileLibrary.cs
```diff
@@ -257,7 +265,9 @@ await FindFiles(true, token)
                         if (response.Any() && response.First().DownloadUrl != null)
                         {
                             var uri = response.First().DownloadUrl;
-                            var filename = Path.Combine(_modelFolder, $"{file.CdfId}.{file.GetExtension()}");
+                            var storageFolder = Path.Combine(_modelFolder, $"{file.CdfId}");
```
lprovensi, line 268: I'm worried about backwards compatibility with previous connector versions. For instance, when updating the connector, it will not be able to delete the files saved by the previous version.

lprovensi (commented):
Logic looks solid. But have you tested backward compatibility and what happens when you upgrade a connector to use this solution given that it has already files stored in the old way?

Cognite.Simulator.Utils/FileLibrary.cs
abdullah-cognite: So, in my local testing, if you just delete the `state.db` and `files` folder when you install this update, everything remains fine. If you dont delete the `files` folder but do delete `state.db`, all the previously saved files still remain in the folder, but they wont be used anymore. I have not tested without deletion of `state.db`, but this will be the worst case scenario.

Cognite.Simulator.Utils/FileLibrary.cs
abdullah-cognite: Done

Cognite.Simulator.Utils/FileLibrary.cs
lprovensi: I suggest testing upgrading the connector as the user would. We cannot guarantee that they will delete the state and the files. Typical case is just installing a new version while the previous one is already installed, with files downloaded and state already there.

Cognite.Simulator.Utils/StateUtils.cs
```diff
@@ -47,7 +48,10 @@ await store.DeleteExtractionState(states, tableName, token)
                 .ConfigureAwait(false);
             foreach (var state in states)
             {
-                DeleteLocalFile(state.FilePath);
+                // get directory path from file path 
+                // (file path is the directory path + file name)
+                var dirPath = Path.GetDirectoryName(state.FilePath);
```
lprovensi, line 53: And this is the risky part. If you try to delete a FilePath that is using the previous format, you would actually ask it to delete the whole `files` folder.

abdullah-cognite: Yes, exactly what i meant as the worst case scenario in my previous comment 😅 . However im not sure of what approach to take, should we delete the state.db when we install a new version? or should we handle backward compatibility directly in the code?

lprovensi: What comes to mind is (considering that GAP is not production yet, and you can manually clean up current GAP installations):
- Deal with compatibility in code, with seamless transition.
- Add a new state property, like `state.Directory`. Assign it just for new models that are downloaded to directories.
- Delete the directory only for models that have assigned `state.Directory`, else keep the previous delete logic.
IMO, this keeps the intention of deleting a folder explicit. Not sure if this holds in practice, would need to be tested 😅

abdullah-cognite: This new logic, is now committed, we now maintain a `IsInDirectory` flag in our state db, this flag is only set for files which are saved in directories, its false for everything that was previously stored :)

Cognite.Simulator.Utils/FileLibrary.cs
```diff
@@ -267,7 +287,7 @@ await FindFiles(true, token)
                                 continue;
                             }
                             file.FilePath = filename;
-
+
```
polomani, line 290:
```suggestion
```

Cognite.Simulator.Utils/FileState.cs
```diff
@@ -49,7 +49,22 @@ public string ModelName
         }
 
         private string _source;
-        
+
+        private bool _isInDirectory = false;
+
```
polomani, line 54: > Member '_isInDirectory' is explicitly initialized to its default value (
false is the default anyway

polomani (approved):
LGTM

PR #83, opened 2023-12-15
fix: default values for version & always use api-based simulation runs 

abdullah-cognite:
The API was giving syntactical request errors, with null values.


polomani (approved):
LGTM,
no new release?

polomani:
Pushed a fix for the current tests.
It seems like the simulation runs API is not enabled by default in the heartbeat and that breaks the tests logic

Cognite.Simulator.Utils/ConnectorBase.cs
```diff
@@ -162,7 +162,7 @@ public virtual bool ShouldLicenseCheck()
         /// <returns></returns>
         public virtual bool ApiEnabled()
         {
-            return _config.UseSimulatorsApi;
+            return true;
```
polomani, line 165: @abdullah-cognite fyi

PR #85, opened 2023-12-21
feat: make end of simulation overridable

abdullah-cognite:
This is required for PROSPER to be able to support legacy simulation runs


Cognite.Simulator.Utils/SimulationRunnerBase.cs
```diff
@@ -252,14 +254,18 @@ public async Task Run(CancellationToken token)
...
+                        {
+                            await StoreRunConfiguration(
+                                calcState,
+                                calcObj,
+                                startTime,
+                                e,
+                                token).ConfigureAwait(false);
+                            _logger.LogDebug("Calculation run finished for event {Id}", eventId);
+                            PublishSimulationRunStatus("IDLE", token);
+                        }
```
polomani, line 268: maybe there was a reason we had it run for in finally block, @lprovensi do you remember?

lprovensi: This is mainly to save the run configuration even when the simulation run fails with an error. I think Abdullah's check is valid, but it is such an edge case that probably never got caught. Good catch 😉

Cognite.Simulator.Utils/SimulationRunnerBase.cs
```diff
@@ -530,6 +536,17 @@ async void PublishSimulationRunStatus(string runStatus, CancellationToken token)
...
+
+            await EndSimulationRun(simEv, startTime, modelState, configState, configObj, metadata, token);
+        }
+
+        protected virtual async Task EndSimulationRun(SimulationRunEvent simEv,
+            DateTime startTime,
+            T modelState,
+            U configState,
+            V configObj,
+            Dictionary<string, string> metadata,
```
lprovensi, line 548: When is the metadata added to the event?

abdullah-cognite: So, since this case of adding metadata to the event was only prosper specific, we override this method in SimConnect here: https://github.com/cognitedata/simconnect-dotnet/pull/148/files#diff-d213be91332c1d1077ea1a1182c3fe747ff008d03b2815b681f3716229a85519R49

lprovensi: We can do it as a workaround, but needs to be documented here. Considering that this code is for a library, it is an anti-pattern to have unused parameters in non-abstract methods.
Also consider if adding meta-data is prosper specific, as it could be a generic mechanism to customize the output events for all connectors.

abdullah-cognite: Good feedback, Ive made the `EndSimulationRun` an abstract with only one parameter which is the simulation event, it will be called at the end of a simulation run and any connector that wants to update their event metadata or do anything else can override it

polomani (approved):
LGTM

Cognite.Simulator.Tests/UtilsTests/SimulationRunnerTest.cs
```diff
@@ -388,6 +388,12 @@ public class SampleSimulationRunner :
         {
         }
 
+        protected override async Task EndSimulationRun(SimulationRunEvent simEv,
+            CancellationToken token)
+        {
+
```
polomani, line 394:
```suggestion
await Task.CompletedTask;
```
I haven't tried this but might help suppressing the warning

PR #89, opened 2024-01-30
[WIP] Routines api scheduler fix

polomani:
potentially makes the scheduler routines-API powered as well


abdullah-cognite (approved):
lgtm

PR #90, opened 2024-02-02
[tech dept] Use robust server-side filters

polomani:
- got rid of some of the client-side filtering in favour of the new filters/endpoints on the API
- use a newer version of the SDK (includes some class renames)

### DIFF
--- a/Cognite.Simulator.Utils/ConfigurationLibraryBase.cs
+++ b/Cognite.Simulator.Utils/ConfigurationLibraryBase.cs
@@ -207,7 +207,7 @@ namespace Cognite.Simulator.Utils
                 throw new ArgumentNullException(nameof(config));
             }
 
-            var routineRevision = await CdfSimulatorResources.RetrieveSimulatorModelRevisionsAsync(
+            var routineRevision = await CdfSimulatorResources.RetrieveSimulatorRoutineRevisionsAsync(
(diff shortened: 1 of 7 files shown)


Cognite.Simulator.Utils/ConfigurationLibraryBase.cs
```diff
@@ -279,8 +273,9 @@ private async Task ReadConfigurations(CancellationToken token)
                 {
                     Filter = new SimulatorRoutineRevisionFilter()
                     {
-                        // TODO filter by created time, simulatorExternalIds, simulatorIntegrationExternalIds
+                        // TODO filter by created time, simulatorIntegrationExternalIds
```
sagar-thalwar, line 276: question(non-blocking): Why do we need to filter by created time?

sagar-thalwar (approved):
👍

Cognite.Simulator.Utils/ConfigurationLibraryBase.cs
polomani: we don't really have to, but this is a way to limit the amount of resources needed to be downloaded from the API every time. (i.e. don't have to refetch the revision that was already fetched yesterday)
This is not currently implemented in the new routines logic though...

polomani: It used to be done with the file-powered APIs

PR #92, opened 2024-02-07
Configuration library clean-up

polomani:
Configuration library clean-up

- Don't be dependent on FileLibrary, only use local DB state
- remove unused methods
- rename TyTope to LocalConfigurationFromRoutine, make a default implementation for it
- fix potential bug in the scheduler when multiple versions of the routine available


sagar-thalwar (approved):
Nice work!

abdullah-cognite (approved):

PR #99, opened 2024-02-27
Remove duplicated simulation run sequence writeback params

polomani:
This removes simulationConfiguration sequence entirely.

We don't need the sequence because most of these params exist on either the simulation run itself or in the routineRevision/modelRevision (can be obtained by using routineRevisionExternalId or modelRevisionExternalId from the run).

We still lack validationStart, validationEnd, samplingStart, and samplingEnd on the simulation run, but if we move validation logic to the backend we don't need them either.
Since we remove the sequence, we also don't need to write runConfigurationSequence, runConfigurationRowStart, runConfigurationRowEnd to the event anymore.

Also removed some code paths where legacy CDF events been used.


abdullah-cognite (approved):
LGTM!

PR #104, opened 2024-03-08
fix integrationExternalId for multiple simulators

abdullah-cognite:
This will allow us to connect to multiple simulators in cases like SimConnect


Cognite.Simulator.Utils/ConnectorBase.cs
```diff
@@ -185,43 +185,50 @@ protected async Task EnsureSimulatorIntegrationsExist(CancellationToken token)
                     token).ConfigureAwait(false);
                 var integrations = integrationRes.Items;
                 var connectorName = GetConnectorName();
-                foreach (var simulator in Simulators)
+                for (var index = 0; index < Simulators.Count; index++)
                 {
-                    var simulatorName = simulator.Name;
+                    var simulator = Simulators[index];
+
+                    // First simulator does not need a suffix (currently used for Simconnect only)
```
polomani, line 192:
```suggestion
                    // If there are more than one, we add the Simulator name as a prefix
                    // E.g. PETEX connector supports PROSPER and GAP
```

polomani (approved):
good

PR #105, opened 2024-03-08
Fix simulation run scope for remote logs

polomani:
This adds a scope on a higher level so we can catch all the errors in the log related to a specific run (e.g. inputs parsing)

now this is possible:
Screenshot 2024-03-08 at 13 44 54


Cognite.Simulator.Utils/SimulationRunnerBase.cs
```diff
@@ -188,71 +188,73 @@ public async Task Run(CancellationToken token)
...
-                        if (calcState == null || calcObj == null || calcObj.Connector != _connectorConfig.GetConnectorName())
+                    using (LogContext.PushProperty("LogId", e.Run.LogId)) {
+                        try
                         {
-                            _logger.LogError("Skip simulation run that belongs to another connector: {Id} {Connector}",
-                               eventId,
-                               calcObj?.Connector);
-                            skipped = true;
-                            continue;
-                        }
```
polomani, line 208: most of this is just an indentation change

lpereiracgn (approved):

abdullah-cognite (approved):

PR #107, opened 2024-03-11
Tests, run them against charts-azuredev project

polomani:
- Time series, datapoints, and files are prepopulated by the test itself, same as routines, models, etc
- data sets are not prepopulated (those can't be deleted, so we should rather not create them automatically)
- we run CICD against azude-dev (instead of westeurope-1)
- fixed a potential bug with the unsafe `ToDictionary()` call (maybe that's what caused the model status to be stuck?)
- renamed FileDownloadClient into FileStorageClient as it now supports upload as well

PS: I felt like tests run a bit slower, maybe we want to switch to Bluefield at some point instead (easy)


sagar-thalwar (approved):
LGTM 🚀

PR #120, opened 2024-04-17
fix: dont send logs under id 0 to the api

lpereiracgn:


abdullah-cognite (approved):

Cognite.Simulator.Utils/ScopedRemoteApiSink.cs
```diff
@@ -96,11 +96,13 @@ private async Task SendToRemoteApi(SimulatorsResource client, CancellationToken
...
-                            logData,
-                            token
-                        ).ConfigureAwait(false);
+                        if(log.Key != 0){
+                            await client.UpdateLogsBatch(
+                                log.Key,
+                                logData,
+                                token
+                            ).ConfigureAwait(false);   
+                        }
```
polomani, line 105: can we please not assign 0 and make a value nullable or smth?
defaulting things to 0 make these sort of bugs to happen in the first place

polomani (changes requested):
make the param optional in case it is optional

Cognite.Simulator.Utils/ScopedRemoteApiSink.cs
polomani: the defaulting logic happens somewhere in the library as well I assume

abdullah-cognite (approved):

abdullah-cognite:
@lpereiracgn test needs a tiny fix

polomani:
@lpereiracgn can you explain in what case we have zeros in this part of the code?

abdullah-cognite:
> @lpereiracgn can you explain in what case we have zeros in this part of the code?

in case logId was mistakenly set to an empty string by any other part of the code.

polomani (approved):
LGTM

PR #125, opened 2024-04-26
Updates for model library

polomani:
- refactor the library so it is powered by the model external IDs instead of model names 
- remove file library, this abstraction doesn't make sense anymore since it's only needed for the model library, but due to some recent changes these two classes always need to be used in conjunction anyway. One file makes it easier to read and reason about the logic
- get model revision from the run (instead of reading the latest always). this allows customization of the model/routine version per run
- write model state to ConcurrentDictionary instead of the normal one, this is done to make sure we it can work in two treads (model library one + actual simulation run)
- make sure we don't write to the state from two different threads (per point), only read
- download/parse model upon the run (only in case it's missing in the state)


Cognite.Simulator.Tests/UtilsTests/SimulationRunnerTest.cs
```diff
@@ -174,20 +178,27 @@ bool checkTs
                 var taskList2 = new List<Task> { runner.Run(linkedToken2) };
                 await taskList2.RunAll(linkedTokenSource2).ConfigureAwait(false);
 
+                // Console.WriteLine($"2 ModelRevision: {modelRevision.Id} \n");
```
polomani, line 181: remove comment

Cognite.Simulator.Utils/ModelLibraryBase.cs
```diff
@@ -1,35 +1,79 @@
...
+        /// <summary>
+        /// Dictionary holding the file states. The keys are the ids of the
+        /// CDF simulator model revisions and the values are the state objects of type <typeparamref name="T"/>
+        /// Should only be modified from a single thread to avoid multi-threading issues.
+        /// </summary>
+        public ConcurrentDictionary<string, T> State { get; private set; }
+
+        /// <summary>
+        /// Temporary model states that are used once and then deleted after the run.
+        /// We keep them in memory to avoid saving them to the state store and hence to avoid the multi-threading issues.
```
abdullah-cognite, line 48:
```suggestion
        /// We keep them in memory to avoid saving them to the state store and hence to avoid the multi-threading issues.
        /// This state is used to store temporary model file paths, used only in case a run event is received before the model file has been downloaded. 
```

abdullah-cognite (approved):

PR #132, opened 2024-05-31
Update enum for licence status

lpereiracgn:


Cognite.Simulator.Utils/ConnectorBase.cs
```diff
@@ -290,7 +290,7 @@ protected async Task EnsureSimulatorIntegrationsExist(CancellationToken token)
         {   
             if (init)
             {
-                LastLicenseCheckResult = ShouldLicenseCheck() ? LicenseStatus.NOT_CHECKED_YET : LicenseStatus.DISABLED;
+                LastLicenseCheckResult = ShouldLicenseCheck() ? LicenseStatus.UNKNOWN : LicenseStatus.DISABLED;
```
polomani, line 293: since you are here

```
XML comment has cref attribute 'SimulatorIntegrationSequenceRows' that could not be resolved

```


maybe you could update the doc string above? :)

polomani (approved):

PR #136, opened 2024-06-13
Model library support reparse

polomani:
The model is now getting re-parsed by the connector only if `status=unknown`. 
To change this default behaviour `public virtual bool ShouldProcess` should be used (e.g. for GAP)

Additional changes and refactoring:
- Remove the ModelName from the local storage. This allows us to decrease the number of API requests since we don't have query model resource (only model revision)
- change the default name for the generated time series. now it includes the reference id
- remove some unused enums etc.
- add model revision parsing status message support


Cognite.Simulator.Extensions/DataModel.cs
```diff
@@ -123,77 +100,23 @@ public static class SimulationVariableMetadata
...
-        /// </summary>
-        SimulationRunConfiguration,
-        /// <summary>
-        /// Simulation calculation data type
-        /// </summary>
-        SimulationEvent,
-        /// <summary>
         /// Model boundary condition data type
         /// </summary>
         BoundaryCondition,
```
abdullah-cognite, line 115: check this

abdullah-cognite (approved):

PR #137, opened 2024-06-18
fix: do not use Task.Wait to avoid deadlocks

polomani:
I noticed the Dwsim doesn't recover when it fails somewhere inside ProcessDownloadedFiles. Inspecting the code I found this (the only place where we use `Wait()`. Removing it to see if it helps

https://blog.stephencleary.com/2012/07/dont-block-on-async-code.html


abdullah-cognite:
I guess we can safely close this one ? @polomani

PR #139, opened 2024-06-28
fix: use model file ext from api

abdullah-cognite:
Refactor and making things more generic: There is no need to get the file extension from the code of each connector when we have that information in the API for each simulator.


sagar-thalwar (approved):
LGTM

PR #141, opened 2024-07-05
fix: release COM object on shutdown

abdullah-cognite:
We dont release the COM object when we're done using it, this can lead to unnecessary instances of the license being consumed. This PR fixes that.


Cognite.Simulator.Utils/Automation/AutomationUtils.cs
```diff
@@ -75,6 +75,8 @@ public virtual void Shutdown()
             }
             finally
             {
+                Marshal.ReleaseComObject(Server);
```
evertoncolling, line 78: That was simpler than I anticipated!

abdullah-cognite:
Closed in favor of : https://github.com/cognitedata/dotnet-simulator-utils/pull/142

PR #142, opened 2024-07-08
fix: release COM object on shutdown

abdullah-cognite:
We dont release the COM object when we're done using it, this can lead to unnecessary instances of the license being consumed. This PR fixes that.


Cognite.Simulator.Utils/Automation/AutomationUtils.cs
```diff
@@ -75,6 +82,10 @@ public virtual void Shutdown()
             }
             finally
             {
+                if (_releaseComObjectByDefault) {
+                    Marshal.ReleaseComObject(Server);
```
polomani, line 86: this is already of the utils

PR #144, opened 2024-07-11
Refactor sampling logic to enable the case where data sampling is disabled

evertoncolling:
This PR introduces the following changes:
- Modify a helper class called `SamplingRange` and rename it to `SamplingConfiguration` -> this object holds the state of whether we should perform data point queries with start, end, aggregates and granularity or if we should fetch data points using latest before.
- Modity the DataExtensions to handle the two cases to fetch data points.
- Adapt the SimulationRunner to properly handle both cases.
- Fixed current tests.

What's missing:
- Update .NET SDK, making `dataSampling` fields nullable (except for `enabled`) -> then updating this PR with null checks on other parts that rely on granularity and aggregates.
- Adding more unit tests covering the new logic


Cognite.Simulator.Extensions/DataPointsExtensions.cs
```diff
@@ -42,6 +42,10 @@ CancellationToken token
             // If the start time is specified, we sample data points with aggregates
             if (samplingConfiguration.Start.HasValue)
             {
+                if (granularity == null)
```
abdullah-cognite, line 45: change to granularity.HasValue

Cognite.Simulator.Extensions/DataPointsExtensions.cs
```diff
@@ -53,7 +57,7 @@ CancellationToken token
                                 Aggregates = new[] {aggregate.AsString()},
                                 Start = $"{samplingConfiguration.Start.Value}",
                                 End = $"{samplingConfiguration.End + 1}", // Add 1 because end is exclusive
-                                Granularity = MinutesToGranularity(granularity),
+                                Granularity = MinutesToGranularity((int)granularity),
```
abdullah-cognite, line 60: change to granularity.Value

Cognite.Simulator.Tests/ExtensionsTests/DataPointsTest.cs
```diff
@@ -89,6 +90,21 @@ public async Task TestGetSample()
...
             Assert.Equal(1, values2[0]);
+
+
+            // DataSampling disabled range
+            var currentTimeEpoch = DateTimeOffset.UtcNow.ToUnixTimeMilliseconds();
+            var disabledDataSampling = new SamplingConfiguration( end: currentTimeEpoch );
+            var (timestamps3, values3) = await dataPoints.GetSample(
+                "SimConnect-IntegrationTests-OnOffValues",
+                Extensions.DataPointAggregate.StepInterpolation,
+                1,
```
abdullah-cognite, line 101: should be null

Cognite.Simulator.Utils/SimulationRunnerBase.cs
```diff
@@ -368,34 +372,47 @@ async void PublishSimulationRunStatus(ConnectorStatus status, CancellationToken
...
-                }
-                // Determine the validation end time
-                if (runItem.Run.RunTime.HasValue)
+                    // Run validation and return a sampling range
+                    var samplingRange = await SimulationUtils.RunSteadyStateAndLogicalCheck(
+                        _cdfDataPoints,
+                        configObj,
+                        validationEnd,
+                        token).ConfigureAwait(false);
+
```
polomani, line 393: @evertoncolling do I understand correctly that SteadyState check and LogicalCheck should only be run if the DataSampling is enabled? Should we in that case forbid enabling these two if DataSampling is disabled? I don't think it happens now on the API validation step

polomani: Another API change is to make this aggregate field optional when data sampling is disabled, right?
That might be the best to validate in the API as well.

Screenshot 2024-08-15 at 12 24 27

Was there something else on API side, do you remember?

evertoncolling: I think you caught everything. In any case let me rewrite it here, to double check:
- Data Sampling Disabled
  - No LC or SSD is performed (they can only be disabled)
  - All datapoints are sampled using latest (before runTime)
  - Aggregates (on timeseries inputs) should be null
- Data Sampling Enabled
  - LC and SSD can be either disabled or enabled
  - All datapoints are sampled using aggregates
  - Sampling Window cannot be null
  - If both LC and SSD are enabled, Validation Window should be null (as it has no effect)
  - Granularity should not be null
  - Aggregates (on timeseries inputs) should not be null

polomani (approved):
I think this is good now

abdullah-cognite (approved):

PR #145, opened 2024-07-16
feat: support timeseries inputs overrides

polomani:
Move the data point sample pulling logic into `LoadTimeseriesSimulationInput()` method and try to read the input value from the overrides first. This makes time series values overrideable with constant values directly on the simulation run.

API changes: https://github.com/cognitedata/infrastructure/pull/12935


abdullah-cognite (approved):
Nice

PR #150, opened 2024-07-30
[POFSP-628] fix remote logger + refactor

abdullah-cognite:
Two main changes here:
1. use a proper dependency injection to create a ScopedRemoteApiSink with a LoggerConfig injected in automatically
2. Move the API logger config to be a part of the logger config instead of the Connector config.
For that, we had to fork a BaseConfig from extractor utils and add our own type of logger config (remote one).

e.g.

```
logger:
  remote:
    level: ""information""
    enabled: true
cognite:
  project:  {project}
  host: {host}
 ```

instead of

```
cognite:
  project:  {project}
  host: {host}
  api-logger:
    level: ""Information""
 ```

Also, there was a bug where the minimum level was not possible to override via the config, now that is fixed.

Fixes the problem here: https://cognitedata.slack.com/archives/C050CRG9ZK4/p1722247823875149


Cognite.Simulator.Utils/ScopedRemoteApiSink.cs
```diff
@@ -18,11 +19,46 @@ namespace Cognite.Simulator.Utils {
...
+                case "debug":
+                    return LogEventLevel.Debug;
+                case "error":
+                    return LogEventLevel.Error;
+                case "information":
+                    return LogEventLevel.Information;
+                case "warning":
+                    return LogEventLevel.Warning;
+                default:
+                    throw new ArgumentException($"Unknown log level: {level}");
```
lpereiracgn, line 59: Check the parsing logic in extractors for setting the log level

polomani: done

lpereiracgn (approved):

PR #155, opened 2024-08-09
Add a simple COM-based simulator tutorial

polomani:


docfx_project/docs/com-connection.md
```diff
@@ -0,0 +1,123 @@
+# Implement COM connection
+
+To be able to run simulations using the Cognite simulator integration, we need to be able to connect to a simulator using `COM` interface.
+In this example, we will use Excel a simple "simulator" and will try to read its version number.
```
abdullah-cognite, line 4:
```suggestion
In this example, we will use Excel as a simple "simulator" and will try to read its version number.
```

docfx_project/docs/com-connection.md
```diff
@@ -0,0 +1,123 @@
...
+    }
+    
+    public static async Task RunStandalone() {
+        Init();
+        await DefaultConnectorRuntime<NewSimAutomationConfig, DefaultModelFilestate, DefaultModelFileStatePoco>.RunStandalone().ConfigureAwait(false);
+    }
+}
+```
+
+### Extend the NewSimClient class to inherit from AutomationClient
```
abdullah-cognite, line 52:
```suggestion
### Extend the NewSimClient class to inherit from the AutomationClient class
```

docfx_project/docs/com-connection.md
```diff
@@ -0,0 +1,123 @@
...
+            {
+                Initialize();
+                _version = Server.Version;
+            }
+            finally
+            {
+                Shutdown();
+            }
+        }
+    }
```
abdullah-cognite, line 97: Lets use semaphoreSlim since its more compatible with Threads statements 

````csharp
private readonly SemaphoreSlim semaphore = new SemaphoreSlim(1, 1);
private string _version = "N/A";

public NewSimClient(ILogger<NewSimClient> logger, DefaultConfig<NewSimAutomationConfig> config)
        : base(logger, config.Automation)
{
    semaphore.Wait();  // Acquire the lock
    try
    {
        Initialize();
        _version = Server.Version;
    }
    finally
    {
        Shutdown();
        semaphore.Release();  // Release the lock
    }
}

File: docfx_project/docs/create-connector.md [hunk 4]
```diff
@@ -0,0 +1,225 @@
...
+        client-id: ${COGNITE_CLIENT_ID}
+        # Client secret
+        secret: ${COGNITE_CLIENT_SECRET}
+        # List of resource scopes, ex:
+        # scopes:
+        #   - https://api.cognitedata.com/.default
+        scopes:
+          - ${COGNITE_SCOPE}
+
+
```
abdullah-cognite, line 48: suggestion: we could mention the `automation` section as well in the config + have a part of the tutorial where the connector uses something from the AutomationConfig, so users know they can define custom stuff in the Config.

docfx_project/docs/create-connector.md
```diff
@@ -0,0 +1,225 @@
...
+```
+
+This file contains the configuration needed to connect to the Cognite Data Fusion project, define the target data set id, and set the connector name.
+
+Make sure to populate the environment variables with the correct values. Alternatively, you can hardcode the values in the configuration file for the development environment.
+
+
+### Create a simulator definition
+
+Now we to have to define a contract between the simulator and the Cognite Data Fusion platform. This contract is defined in the API as a `Simulator` object.
```
abdullah-cognite, line 65: I like the `contract` term 🚀

docfx_project/docs/implement-routine.md
```diff
@@ -0,0 +1,240 @@
...
+    }
+
+    public override void RunCommand(Dictionary<string, string> arguments)
+    {
+        // No implementation needed for this simulator
+    }
+}
+```
+This class will be used to perform the simulation.
+The `SetInput` method is used to set the input values for the simulation. The `GetOutput` method is used to get the output values from the simulation. The `RunCommand` method is used to run commands in the simulation, but it is not needed for this simulator as the results are calculated immediatelly on the worksheet.
```
abdullah-cognite, line 92:
```suggestion
The `SetInput` method is used to set the input values for the simulation. The `GetOutput` method is used to get the output values from the simulation. The `RunCommand` method is used to run commands in the simulation, but it is not needed for this simulator as the results are calculated immediately on the worksheet.
```

abdullah-cognite (commented):
nice! loved following through the tutorial, left a few small comments

docfx_project/docs/create-connector.md
polomani: I think this would require another tutorial (as an another MD page), so maybe we could give more complex examples later?
Currently, this is rather simple tutorial, I agree.
I just can't really come up with any custom stuff idea for the Excel currently...

sowmyaraj92 (approved):
Reviewed and approved. 🚀

abdullah-cognite (approved):

PR #157, opened 2024-08-19
try to make tests faster

polomani:
- make seed data a class
- do not wait X seconds, but rather listen to the sim run status


abdullah-cognite:
Can we close this one too @polomani

polomani:
yeah this is too old

PR #178, opened 2024-09-19
add a small test for scheduler

abdullah-cognite:
I dont really know where Im going with this, I just created this to see if we can catch the issue with incorrect cron expressions on the connector. Seems like incorrect expressions should already be caught by the NCrontab library, but for some reason we still ran into an issue by providing the wrong expression. 🤔

### DIFF
--- a/Cognite.Simulator.Utils/SimulationScheduler.cs
+++ b/Cognite.Simulator.Utils/SimulationScheduler.cs
@@ -73,6 +73,19 @@ namespace Cognite.Simulator.Utils
         /// Routine revision.
         /// </summary>
         public V RoutineRevision { get; set; }
+
+        /// <summary>
+        /// Set schedule
+        /// </summary>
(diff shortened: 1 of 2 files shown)


Cognite.Simulator.Utils/SimulationScheduler.cs
```diff
@@ -73,6 +73,19 @@ public class ScheduledJob<V> where V : SimulatorRoutineRevision
...
+        /// </summary>
+        /// <param name="schedule">Crontab schedule</param>
+        /// <returns>Self</returns>
+        /// <exception cref="CrontabException">If the schedule is invalid</exception>
+        /// 
+        public ScheduledJob<V> SetSchedule(string CronExpression)
+        {
+            Schedule = CrontabSchedule.Parse(CronExpression);
+            return this;
+        }
```
polomani, line 88: why not having this in the constructor?

PR #181, opened 2024-10-02
tests: use a new ts external id on each run

polomani:
this tries to assign new timeseries ids for inputs/outputs every time the test runs. should help with running tests in parallel.

we currently have a lot of error like: `No output time series were created [SimConnect-IntegrationTests-OT1-Output]`

### DIFF
--- a/Cognite.Simulator.Tests/ExtensionsTests/DataPointsTest.cs
+++ b/Cognite.Simulator.Tests/ExtensionsTests/DataPointsTest.cs
@@ -49,13 +49,15 @@ namespace Cognite.Simulator.Tests.ExtensionsTests
             var cdf = provider.GetRequiredService<Client>();
             var dataPoints = cdf.DataPoints;
 
+            await SeedData.GetOrCreateTestTimeseries(cdf).ConfigureAwait(false);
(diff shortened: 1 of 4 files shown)


Cognite.Simulator.Tests/ExtensionsTests/DataPointsTest.cs
```diff
@@ -49,13 +49,15 @@ public async Task TestGetSample()
...
+            await SeedData.GetOrCreateTestTimeseries(cdf).ConfigureAwait(false);
+
             // Tests assume the given time series exists in the test project and
             // that it contains data points in this time range
             var samplingConfiguration = new SamplingConfiguration(start: 1631294940000, end: 1631296740000);
             
             // Test max aggregation. A single data point should be returned
             var (timestamps, values) = await dataPoints.GetSample(
-                "SimConnect-IntegrationTests-OnOffValues",
+                "utils-tests-OnOffValues",
```
polomani, line 60: this is just a rename so we don't mix the test cases with the actual `SimConnect`

lpereiracgn (approved):

PR #184, opened 2024-10-02
fix: Upgrade to .NET 8

lpereiracgn:
Updates all targets to use .NET 8 and C# v12 as the language version.


Cognite.DataProcessing/Cognite.DataProcessing.csproj
```diff
@@ -1,7 +1,8 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFramework>netstandard2.0</TargetFramework>
```
polomani, line 4: I am thinking of keeping netstandard2.0 as a target for the library itself and using `net8.0` for the other projects (tests, example connectors).

This is because we don't know where the library might be used in the future, and the more compatible it is the better. netstandard2.0 is just an interface, not the implementation, so we still getting some of the .net8 performance boosts if we run it on it

polomani (approved):
LGTM

PR #187, opened 2024-10-09
Fix connector getting stuck on API error

polomani:
A few things happened here

1. The major one was passing the wrong cancellation token into the `taskList.Add(_pipeline.PipelineUpdate(token));`.
Because of the wrong token, the connector was failing to perform a "soft" restart, i.e. rerun the init sequence + start a new loop.
This is now fixed as well and we have a runtime test with a mocked HttpClientFactory that is able to generate failing HTTP responses, so we can test things like this more going forward.
3. `await _pipeline.Init(_config.Simulator, token).ConfigureAwait(false);` Was the last step in the initialization sequence which prevents errors being printed to the extraction pipeline on the fail-to-start event of the connector.
4. `var pipeline = provider.GetRequiredService<ExtractionPipeline>();` was called with the wrong scope which also prevented some errors from being reported to the extraction pipeline.
5. Some minor changes in error handling


Cognite.Simulator.Tests/IntegrationTests/ConnectorRuntimeTest.cs
```diff
@@ -93,63 +96,54 @@ private void WriteConfig()
         [Fact]
         public async Task TestConnectorRuntime()
         {
```
polomani, line 98: I barely touched this particular test, this is just a refactor.

Cognite.Simulator.Tests/UtilsTests/ConnectorRuntimeUnitTest.cs
```diff
@@ -0,0 +1,132 @@
...
+            try {
+                await DefaultConnectorRuntime<DefaultAutomationConfig, DefaultModelFilestate, DefaultModelFileStatePoco>.Run(logger, cts.Token).ConfigureAwait(false);
+            } catch (OperationCanceledException) {}
+        
+            // Check the logs, it should first succeed on the startup, then fail and restart
+            VerifyLog(mockedLogger, LogLevel.Information, "Starting the connector...", Times.Once());
+            VerifyLog(mockedLogger, LogLevel.Information, "Connector can reach CDF!", Times.Once());
+            VerifyLog(mockedLogger, LogLevel.Debug, "Updating simulator definition", Times.Once());
+            VerifyLog(mockedLogger, LogLevel.Error, "Request to CDF failed with code 403", Times.Once(), true);
+            VerifyLog(mockedLogger, LogLevel.Warning, "Restarting connector in 5 seconds", Times.AtLeastOnce());
```
lpereiracgn, line 69: Not sharing any review feedback just leaving a comment to appreciate how good this solution looks.

polomani: `VerifyLog` specifically is something @abdullah-cognite made, so I am not taking a credit for that!
I did the HTTP mocking part 😇

Cognite.Simulator.Utils/DefaultClasses/DefaultConnectorRuntime.cs
```diff
@@ -192,7 +192,7 @@ public static async Task Run(ILogger defaultLogger, CancellationToken token)
         {
             using (var scope = provider.CreateScope())
             {
-                var pipeline = provider.GetRequiredService<ExtractionPipeline>();
+                var pipeline = scope.ServiceProvider.GetRequiredService<ExtractionPipeline>();
```
abdullah-cognite, line 195: Ah! nice! good catch!

Cognite.Simulator.Tests/UtilsTests/TestUtilities.cs
```diff
@@ -0,0 +1,183 @@
...
+
+                foreach (var mapping in endpointMappings)
+                {
+                    if (mapping.Key(uri))
+                    {
+                        var (responseFunc, callCount, maxCalls) = mapping.Value;
+                        if (maxCalls.HasValue) {
+                            if (callCount < maxCalls)
+                            {
+                                endpointMappings[mapping.Key] = (responseFunc, callCount + 1, maxCalls);
```
abdullah-cognite, line 115: We could make this thread-safe when tests run concurrently by making this a `ConcurrentDictionary`

Cognite.Simulator.Tests/UtilsTests/TestUtilities.cs
```diff
@@ -0,0 +1,183 @@
...
+                    throw new ArgumentNullException(nameof(uri));
+                }
+
+                foreach (var mapping in endpointMappings)
+                {
+                    if (mapping.Key(uri))
+                    {
+                        var (responseFunc, callCount, maxCalls) = mapping.Value;
+                        if (maxCalls.HasValue) {
+                            if (callCount < maxCalls)
```
abdullah-cognite, line 113: We could probably refactor this a bit more to something like :

```
if (maxCalls.HasValue && callCount >= maxCalls)
{
    return ForbiddenResponse;
}

File: Cognite.Simulator.Tests/UtilsTests/TestUtilities.cs (hunk 4 above)
Comment (polomani) [reply Cognite.Simulator.Tests/UtilsTests/TestUtilities.cs:115]: this is used for the tests (we run them in one thread, because there will be more issues if we don't)

Comment (polomani) [reply Cognite.Simulator.Tests/UtilsTests/TestUtilities.cs:115]: yeah, let's try that

**Comment by abdullah-cognite** [comment]
Overall looks good, one suggestion, since we are beta now, lets introduce a CHANGELOG.md file, so consumers of this library know what changed in each version

File: Cognite.Simulator.Tests/IntegrationTests/ConnectorRuntimeTest.cs [hunk 6]
```diff
@@ -93,63 +96,54 @@ private void WriteConfig()
...
-            var services = new ServiceCollection();
-            services.AddSingleton<DefaultConfig<AutomationConfig>>();
```
abdullah-cognite, line 97: Since this test still works without the addsingleton line , I assume removing this is fine

polomani: yes, it was an incorrect place to inject the dependency. it didn't affect the test in any way because this service collection is not used by the runtime of the connector

Cognite.Simulator.Tests/UtilsTests/TestUtilities.cs
polomani: ok!

abdullah-cognite (approved):

PR #188, opened 2024-10-09
fix: use utc time in the scheduler [POFSP-800]

polomani:
just making sure we are using Utc everywhere. right now this is the main hypothesis around the issues on the project

### DIFF
--- a/Cognite.Simulator.Utils/SimulationScheduler.cs
+++ b/Cognite.Simulator.Utils/SimulationScheduler.cs
@@ -1,14 +1,17 @@
-﻿using Cognite.Extractor.Utils;
-using CogniteSdk.Alpha;
-using Microsoft.Extensions.Logging;
-using System;
+﻿using System;
 using System.Collections.Generic;
 using System.Linq;
 using System.Threading;
 using System.Threading.Tasks;
+
+using Microsoft.Extensions.Logging;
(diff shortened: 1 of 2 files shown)


evertoncolling (approved):
Good catch!

Cognite.Simulator.Utils/SimulationScheduler.cs
```diff
@@ -225,14 +225,15 @@ private async Task RunJob(ScheduledJob<V> job, CancellationToken mainToken)
             while (!mainToken.IsCancellationRequested || !job.TokenSource.Token.IsCancellationRequested)
             {
                 var routineRev = job.RoutineRevision;
-                var nextOccurrence = job.Schedule.GetNextOccurrence(DateTime.Now);
-                var delay = nextOccurrence - DateTime.Now;
+                var now = DateTime.UtcNow;
```
sagar-thalwar, line 228: Nitpick: Is it better to create a util function `DateTime.UtcNow` instead of using the same logic at different places?

polomani: Encapsulated this into a separate function now, but yeah, I agree, we might want to consider a separate wrapper around DateTime

sagar-thalwar (approved):
LGTM

PR #196, opened 2024-10-24
feat: override log severity, adopt future breaking changes

polomani:
- override min log severity level per simulation run
- adopt `IncludeAllFields` property and smaller response pages on routine revisions list
- minor rename in prop filter


Cognite.Simulator.Utils/Logger.cs
```diff
@@ -16,8 +23,38 @@ namespace Cognite.Simulator.Utils
     /// Loggers are created according to a <see cref="LoggerConfig"/> configuration object.
     /// Log messages contain UTC timestamps.
     /// </summary>
-    public static class SimulatorLoggingUtils 
+    public static class SimulatorLoggingUtils
     {
+        /// <summary>
+        /// Used to enrish log events with the simulator log id and min severity.
```
abdullah-cognite, line 29:
```suggestion
        /// Used to enrich log events with the simulator log id and min severity.
```

abdullah-cognite (approved):
Looks okay, so now with every new LogContext we ware making one API call too 🤔

polomani:
yes, but only on simulation run endpoint

PR #219, opened 2024-12-06
feat: support large files

polomani:
- support files up to 8 gb (theoretically)
- retry download a few times, instead of forever
- do not allow large (512+) MB files to be downloaded&parsed "on a fly"
- allow the file to be re-downloaded when re-parsing triggered and file was never downloaded before


Cognite.Simulator.Utils/FileState.cs
```diff
@@ -222,6 +222,21 @@ public int Version
             }
         }
 
+        private int _downloadAttempts;
+
+        /// <summary>
+        /// Download attempts
+        /// <summary/>
```
polomani, line 229:
```suggestion
        /// </summary>
```

Cognite.Simulator.Utils/FileStorageClient.cs
```diff
@@ -29,38 +33,65 @@ public FileStorageClient(HttpClient client, ILogger<FileStorageClient> logger)
...
-                var response = await _client.GetAsync(uri)
+                var response = await _client.GetAsync(uri, HttpCompletionOption.ResponseHeadersRead)
                     .ConfigureAwait(false);
                 if (!response.IsSuccessStatusCode)
                 {
                     _logger.LogWarning("Failed to download the file: {Message}", response.ReasonPhrase);
                     return false;
                 }
 
+                if (response.Content.Headers.ContentLength > MaxFileDownloadSize)
```
abdullah-cognite, line 52: Could it be the case that `ContentLength`  is null in the headers? should we add a null check here?

Cognite.Simulator.Utils/FileStorageClient.cs
```diff
@@ -29,38 +33,65 @@ public FileStorageClient(HttpClient client, ILogger<FileStorageClient> logger)
...
+                    _logger.LogWarning("File size exceeds the maximum allowed size: {Size} bytes", MaxFileDownloadSize);
+                    return false;
+                }
+
+                if (failLargeFiles && response.Content.Headers.ContentLength > LargeFileSize)
+                {
+                    throw new ConnectorException(
+                        $"The file is larger than {LargeFileSize / 1024 / 1024} MB and needs to be pre-downloaded before any operations can be performed." +
+                        "Connector hasn't been able to pre-download the file yet. " +
+                        "Please wait for the file to be downloaded and try again."
```
abdullah-cognite, line 63: The error says `Please wait for the file to be downloaded and try again.` But actually if `failLargeFiles` is true, the connector would keep throwing the same exception on retrying? so should we remove the `try again` part

Cognite.Simulator.Utils/FileStorageClient.cs
```diff
@@ -29,38 +33,65 @@ public FileStorageClient(HttpClient client, ILogger<FileStorageClient> logger)
...
+                }
+
                 if (File.Exists(filePath))
                 {
                     File.Delete(filePath);
                 }
-                using (var fs = new FileStream(filePath, FileMode.CreateNew))
+
+                using (var responseStream = await response.Content.ReadAsStreamAsync().ConfigureAwait(false))
+                using (var fileStream = new FileStream(filePath, FileMode.CreateNew, FileAccess.Write, FileShare.None, bufferSize: DefaultBufferSize, useAsync: true))
```
abdullah-cognite, line 73: why did we add these new parameters? `FileAccess.Write, FileShare.None,`

Is it because of `useAsync`?

Cognite.Simulator.Utils/FileStorageClient.cs
```diff
@@ -29,38 +33,65 @@ public FileStorageClient(HttpClient client, ILogger<FileStorageClient> logger)
...
                 }
             }
             catch (HttpRequestException e)
             {
                 // File cannot be downloaded, skip for now and try again later
-                _logger.LogError("Failed to download the file: {Message}", e.Message);
-                return false;
+                _logger.LogError("Failed to download the file {uri}: {Message}", uri, e.Message);
+            }
+            catch (IOException e)
```
abdullah-cognite, line 90: Should we also catch generic Exceptions here?

Cognite.Simulator.Utils/ModelLibraryBase.cs
```diff
@@ -393,16 +393,24 @@ await _cdfSimulatorResources.UpdateSimulatorModelRevisionParsingStatus(
...
         private void UpdateModelParsingInfo(T modelState, SimulatorModelRevision modelRevision)
         {
-            if (modelState.ParsingInfo == null || modelState.ParsingInfo.LastUpdatedTime < modelRevision.LastUpdatedTime)
+            var status = modelRevision.Status;
+            var needsUpdate = modelState.UpdatedTime < modelRevision.LastUpdatedTime;
+            if (needsUpdate && status == SimulatorModelRevisionStatus.unknown)
+            {
+                _logger.LogDebug("Resetting download attempts counter for {ModelExtid} due to unknown status",
+                    modelState.ModelExternalId,
+                    modelState.ParsingInfo,
```
abdullah-cognite, line 402: is the logger able to properly print the `ParsingInfo `object?

Cognite.Simulator.Utils/ModelLibraryBase.cs
```diff
@@ -569,11 +578,11 @@ private async Task<bool> DownloadFileAsync(T modelState, bool isTemporary = fals
                     modelState.IsInDirectory = true;
                     
                     bool downloaded = await _downloadClient
-                        .DownloadFileAsync(uri, filename)
+                        .DownloadFileAsync(uri, filename, failLargeFiles: isTemporary)
```
abdullah-cognite, line 581: Nice 👍

Cognite.Simulator.Tests/UtilsTests/ModelLibraryTest.cs
```diff
@@ -208,7 +209,7 @@ await lib.GetRunTasks(linkedTokenSource.Token)
                     Assert.Equal(SimulatorModelRevisionStatus.failure, modelInState.ParsingInfo.Status);
                     Assert.True(modelInState.ParsingInfo.Error);
                     // last updated time should not change
-                    Assert.Equal(revision.LastUpdatedTime, modelInState.ParsingInfo.LastUpdatedTime);
+                    Assert.Equal(revision.LastUpdatedTime, modelInState.UpdatedTime);
```
abdullah-cognite, line 212: Can we reduce the `LargeFileSize` externally to something very small like 1MB and test that code with a generated file?

Cognite.Simulator.Utils/FileStorageClient.cs
polomani: no, this only fails when the file isn't downloaded beforehand and there is a simulation run that tries to use it.
in that case we normally try to download i on a fly, remember?
however, this doesn't make sense to keep when files are big, because they take time to be downloaded, so the simulations will get stuck for a long time. so better to fail it quick and let the library download it later

Cognite.Simulator.Utils/FileStorageClient.cs
polomani: because of the `bufferSize` param

Cognite.Simulator.Utils/FileStorageClient.cs
polomani: it's caught one level above on the callee in `ModelLibraryBase.cs`

Cognite.Simulator.Utils/ModelLibraryBase.cs
polomani: this is not used, and you are right

Cognite.Simulator.Tests/UtilsTests/ModelLibraryTest.cs
polomani: it's a bit of work, but I'll try to see what we can do with a fresh eyes on Monday then

Cognite.Simulator.Utils/FileStorageClient.cs
polomani: ref this `.DownloadFileAsync(uri, filename, failLargeFiles: isTemporary)`

Cognite.Simulator.Utils/FileStorageClient.cs
polomani: it's a nullable long. so yeah, we should add a check

Cognite.Simulator.Tests/UtilsTests/ModelLibraryTest.cs
abdullah-cognite: looks good

abdullah-cognite (approved):
lgtm

PR #235, opened 2025-02-12
fix(scheduler): change to single task based scheduler

abdullah-cognite:
Current scheduler has a dotnet `Task` for each routine revision. This leads to problems when a large number of RoutineRevisions are scheduled at the exact same time every hour. This PR changes the approach so we dont create Tasks at all, rather have just one `Task` that keeps waking up every minute and goes through all the routines in the state to decide which needs to be scheduled based on their cron expression.


Cognite.Simulator.Utils/SimulationScheduler.cs
```diff
@@ -148,70 +165,86 @@ await Task.Run(async () =>
...
+                            lastRun = DateTime.MinValue;
                         }
 
-                        if (!scheduledJobs.TryGetValue(routineRev.RoutineExternalId, out var existingJob))
+                        // Compute the next scheduled occurrence from the last run time.
+                        var nextOccurrence = GetNextOccurrence(schedule, lastRun);
+
+                        // If the next occurrence is less or equal than the current UTC time (plus tolerance)
+                        // and is greater than the last run time, trigger the simulation run.
+                        if (nextOccurrence <= DateTime.UtcNow.Add(tolerance) && nextOccurrence > lastRun)
```
polomani, line 193:
```suggestion
                        if (nextOccurrence >= DateTime.UtcNow && nextOccurrence <= DateTime.UtcNow.Add(tolerance) && nextOccurrence > lastRun)
```
otherwise it might fire before the actual schedule (we can't allow that)

Cognite.Simulator.Utils/SimulationScheduler.cs
```diff
@@ -43,50 +42,25 @@ public async Task Delay(TimeSpan delay, CancellationToken token)
 
     /// <summary>
     /// Represents a scheduled job for simulation.
+    /// (No longer used since scheduling is now done in a single loop.)
```
polomani, line 45: why are you keeping it then?

Cognite.Simulator.Utils/SimulationScheduler.cs
```diff
@@ -148,70 +165,86 @@ await Task.Run(async () =>
                             continue;
                         }
 
-                        // Check if the job already exists and if it should be cancelled due 
-                        // to a new configuration on the API
-                        if (scheduledJobs.TryGetValue(routineRev.RoutineExternalId, out var job))
+                        // Parse the cron schedule.
+                        CrontabSchedule schedule;
```
polomani, line 169: maybe put this into the map as well, so we don't recompute it each time?

polomani: could probably repurpose `ScheduledJob` for this

PR #237, opened 2025-02-17
attempt to address stability issues

abdullah-cognite:
* **Breaking** added a method on the `ISimulatorClient` called `TestConnection`. From this build onwards there will be no need to call the `TestConnection` function from the `ConfigureServices` section of each simulator.
* Wrapped some more areas of the code with Try-Catch statements to enhance stability.
* Added support for backoff based retries to API calls.


Cognite.Simulator.Utils/RetryPolicy.cs
```diff
@@ -0,0 +1,71 @@
...
+        /// </summary>
+        /// <param name="services">The service collection to add the HTTP client to.</param>
+        /// <returns>The updated service collection.</returns>
+        public static IServiceCollection ConfigureCustomHttpClientWithRetryPolicy(this IServiceCollection services)
+        {
+            services.AddHttpClient<Client.Builder>()
+                .AddPolicyHandler((serviceProvider, _) =>
+                {
+                    return HttpPolicyExtensions
+                        .HandleTransientHttpError()
```
sagar-thalwar, line 54: (blocking) : Can(should?) we ensure that 5xx errors are not retried?

Cognite.Simulator.Utils/RetryPolicy.cs
```diff
@@ -0,0 +1,71 @@
...
+        public static Action<Exception, TimeSpan, int, Context, IServiceProvider> OnRetry { get; set; } = (exception, timeSpan, retryCount, context, serviceProvider) => { 
+            try
+            {
+                var logger = serviceProvider.GetService<ILogger<Client.Builder>>();
+                if (logger != null)
+                {
+                    logger.LogWarning(
+                        "Retry {RetryCount} of {MaxRetries} after {Delay}s. Error: {Error}",
+                        retryCount,
+                        11,
```
sagar-thalwar, line 27: (blocking): I think we should make this configurable instead of hardcoding to 11 retires. (I am not talking about the connector config)

sagar-thalwar (commented):
Few comments. I am still taking a look at the tests.

Cognite.Simulator.Utils/RetryPolicy.cs
```diff
@@ -0,0 +1,71 @@
...
+    /// <summary>
+    /// Configuration for HTTP client policies.
+    /// </summary>
+    public static class HttpClientPolicyConfiguration
+    {
+        
+        /// <summary>
+        /// Gets or sets the action to execute when a transient HTTP error occurs.
+        /// </summary>
+        public static Action<Exception, TimeSpan, int, Context, IServiceProvider> OnRetry { get; set; } = (exception, timeSpan, retryCount, context, serviceProvider) => {
```
abdullah-cognite, line 18: There might not be need of this, we might able to use the extractor utils to do this: https://github.com/cognitedata/dotnet-extractor-utils/blob/67442743d275d1e8393c79a49275b990593285a7/Cognite.Extensions/CogniteUtils.cs#L551C57-L551C71

Cognite.Simulator.Utils/RetryPolicy.cs
abdullah-cognite: Ive updated the code to use this directly from the extractor utils (it is config driven from there). The default value was quite low in the extractor utils (`5`), i have raised it to 11 for us.

https://docs.cognite.com/cdf/integration/guides/simulators/connectors/

Cognite.Simulator.Utils/RetryPolicy.cs
abdullah-cognite: Only these status codes should be retried: https://github.com/cognitedata/dotnet-extractor-utils/blob/67442743d275d1e8393c79a49275b990593285a7/Cognite.Extensions/CogniteUtils.cs#L551C57-L551C71

sagar-thalwar (approved):

Cognite.Simulator.Utils/Configuration.cs
```diff
@@ -74,6 +74,11 @@ public override void GenerateDefaults()
                 Cognite = new CogniteConfig();
             }
 
+            if (Cognite.CdfRetries == null) {
+                Cognite.CdfRetries.MaxRetries = 11;
+                Cognite.CdfRetries.MaxDelay = 60 * 1000; //ms
+            }
+
```
polomani, line 81: if `CdfRetries == null`  the next line will cause null pointer exception

PR #238, opened 2025-02-17
feat: cleanup old model revision files

lpereiracgn:
This change introduces a new property, `LastAccessTime`, in the local state for simulator model revisions. Each time a revision is accessed, this property will be updated. Based on the `FilesTTL` value defined in the `config.yaml`, it will check whether the last usage of a file exceeds the specified retention period. If the file has not been accessed within the defined timeframe, it will be deleted.

### DIFF
--- a/Cognite.Simulator.Tests/UtilsTests/ModelLibraryTest.cs
+++ b/Cognite.Simulator.Tests/UtilsTests/ModelLibraryTest.cs
@@ -18,6 +18,7 @@ using CogniteSdk.Alpha;
 
 using Cognite.Simulator.Extensions;
 using Cognite.Simulator.Utils.Automation;
+using Microsoft.Extensions.DependencyInjection.Extensions;
 
 namespace Cognite.Simulator.Tests.UtilsTests
 {
@@ -286,6 +287,7 @@ namespace Cognite.Simulator.Tests.UtilsTests
             services.AddSingleton<ModelParsingInfo>();
             services.AddSingleton<ScopedRemoteApiSink>();
             services.AddSingleton<DefaultConfig<AutomationConfig>>();
+            services.AddSingleton<IDateTimeProvider, SystemDateTimeProvider>();
             StateStoreConfig stateConfig = null;
             using var provider = services.BuildServiceProvider();
 
@@ -417,6 +419,7 @@ namespace Cognite.Simulator.Tests.UtilsTests
             services.AddSingleton<ScopedRemoteApiSink>();
             services.AddSingleton(SeedData.SimulatorCreate);
             services.AddSingleton<DefaultConfig<AutomationConfig>>();
+            services.AddSingleton<IDateTimeProvider, SystemDateTimeProvider>();
             StateStoreConfig stateConfig = null;
             using var provider = services.BuildServiceProvider();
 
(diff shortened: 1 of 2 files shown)


Cognite.Simulator.Utils/ModelLibraryBase.cs
```diff
@@ -241,14 +246,42 @@ public async Task<T> GetModelRevision(string modelRevisionExternalId)
...
             var model = modelVersions.FirstOrDefault();
 
-            if (model != null)
-            {
+            if(model != null){
+                if(model.IsDeleted){
+                    var downloaded = await DownloadFileAsync(model, true).ConfigureAwait(false);
+                    if(downloaded){
+                        model.IsDeleted = false;
+                    }
```
polomani, line 254: if you remove the revision from the state (instead of only deleting the file) you should not need this

Cognite.Simulator.Utils/ModelLibraryBase.cs
```diff
@@ -241,14 +246,42 @@ public async Task<T> GetModelRevision(string modelRevisionExternalId)
...
+        /// <summary>
+        /// If a file has not been accessed for a certain amount of time, it will be deleted.
+        /// The default is 7 days.
+        /// </summary>
+        public void CleanExpiredModelRevisionFiles(){
+            var revisions = _state.Values
+                .Where(s => s.IsInDirectory
+                    && !string.IsNullOrEmpty(s.FilePath)
+                    && s.LastAccessTime.AddDays(_config.FilesTTL) < _dateTimeProvider.UtcNow)
+                .OrderByDescending(s => s.CreatedTime);
```
polomani, line 274: Multiple revisions can reuse one file, so this is unsafe.
also, removing the latest revision for the model is most likely not a good idea.

this code from `VerifyLocalModelState` could be taken as an example (or reused?)

```
var filesInUseMap = _state.Values
                        .Where(f => !string.IsNullOrEmpty(f.FilePath))
                        .ToDictionarySafe(f => f.FilePath, f => true);
                    if (statesToDelete.Any())
                    {
                        _logger.LogWarning("Removing {Num} model versions not found in CDF: {Versions}",
                            statesToDelete.Count,
                            string.Join(", ", statesToDelete.Select(s => s.ModelExternalId + " v" + s.Version)));

                        var statesToDeleteWithFile = statesToDelete
                            .Select(s => (s, !filesInUseMap.ContainsKey(s.FilePath)));

                        await _store
                            .RemoveFileStates(_config.FilesTable, statesToDeleteWithFile, token)
                            .ConfigureAwait(false);
                    }
                    
```

Cognite.Simulator.Utils/ModelLibraryBase.cs
```diff
@@ -196,6 +200,7 @@ await _store.RestoreExtractionState<U, T>(
                     HashSet<string> idsToKeep = new HashSet<string>(_state.Select(s => s.Value.Id));
                     ldbStore.RemoveUnusedState(_config.FilesTable, idsToKeep);
                 }
+                CleanExpiredModelRevisionFiles();
```
polomani, line 203: only on restart? this may never happen if the connector is stable enough

PR #240, opened 2025-03-05
fix: retry init due to intermittent errors on extraction pipeline

polomani:
- Generally extraction pipeline errors should not stop the connector from being initialised / running (current state)
- however, there are cases when the extraction pipeline create/byids call fails on startup due to whatever reasons (the API itself is soon to be replaced with v2, so I don't think it's wise to dig for the root cause)
- This a problem, because some people rely on the ext pipeline notifications (e.g. connector is down), and previously the pipeline would be off if the initialisation failed from the get go. This did cause false notifications.
- this patch allows the ext pipeline to be initialized later on runtime whenever it's needed


Cognite.Simulator.Tests/UtilsTests/ExtPipelineUnitTest.cs
```diff
@@ -0,0 +1,85 @@
...
+
+        [Fact]
+        /// <summary>
+        /// Test the late initialization of the extraction pipeline.
+        /// 1. On the first attempt it tries to get the pipeline, the /extpipes/byids fails with a 400.
+        /// 2. Next try, the /extpipes/byids returns an empty response. So the connector will try to create the pipeline. Create fails with a 400.
+        /// 3. Next try, the /extpipes/byids returns an empty response. The connector will try to create the pipeline again. Create succeeds.
+        /// 4. The connector will then try to notify the pipeline, which should succeed.
+        /// </summary>
+        public async Task TestExtPipineRetryInitRemote()
```
abdullah-cognite, line 41: typo
```suggestion
        public async Task TestExtPipelineRetryInitRemote()
```

Cognite.Simulator.Tests/UtilsTests/TestUtilities.cs
```diff
@@ -101,24 +102,67 @@ public static Func<HttpRequestMessage, CancellationToken, Task<HttpResponseMessa
...
+            }
+
+            public bool Matches(string uri)
+            {
+                return _uriMatcher(uri) && HasMoreCalls();
+            }
+
+            public HttpResponseMessage GetResponse()
+            {
+                _callCount++;
```
abdullah-cognite, line 151: should there be a guard here if we exceed the `maxCalls`?

Cognite.Simulator.Tests/UtilsTests/TestUtilities.cs
```diff
@@ -101,24 +102,67 @@ public static Func<HttpRequestMessage, CancellationToken, Task<HttpResponseMessa
...
-
-                return new HttpResponseMessage(HttpStatusCode.OK) { Content = new StringContent("{\"items\":[]}") };
+                return CreateResponse(HttpStatusCode.NotImplemented, "Not implemented");
             };
         }
 
+        /// <summary>
+        /// A container for easier mocking of endpoints.
+        /// </summary>
+        public class SimpleRequestMocker
```
abdullah-cognite, line 119: Is this going to be thread safe for parallel tests?

Cognite.Simulator.Tests/UtilsTests/TestUtilities.cs
```diff
@@ -101,24 +102,67 @@ public static Func<HttpRequestMessage, CancellationToken, Task<HttpResponseMessa
...
+        {
+            private readonly Func<string, bool> _uriMatcher;
+            private readonly Func<HttpResponseMessage> _responseFunc;
+            private int _callCount = 0;
```
abdullah-cognite, line 123: Would exposing this be useful for our tests?

Cognite.Simulator.Tests/UtilsTests/TestUtilities.cs
polomani: ok!

Cognite.Simulator.Tests/UtilsTests/ExtPipelineUnitTest.cs
```diff
@@ -0,0 +1,85 @@
...
+            // Act
+            await extPipeline.Init(connectorConfig, CancellationToken.None);
+
+            using var tokenSource = new CancellationTokenSource();
+            tokenSource.CancelAfter(TimeSpan.FromSeconds(5));
+
+            try
+            {
+                await extPipeline.PipelineUpdate(tokenSource.Token);
+            } catch (OperationCanceledException) { }
```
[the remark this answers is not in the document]
polomani: ok

Cognite.Simulator.Tests/UtilsTests/TestUtilities.cs
polomani: we use SequentialTestCollection to run them, so we should not worry. (there will most likely be more issues otherwise)

Cognite.Simulator.Tests/UtilsTests/TestUtilities.cs
polomani: I exposed something else now => `AssertCallCount` should be much more useful

Cognite.Simulator.Tests/UtilsTests/TestUtilities.cs
```diff
@@ -101,24 +103,89 @@ public static Func<HttpRequestMessage, CancellationToken, Task<HttpResponseMessa
...
+            public void AssertCallCount()
+            {
+                Assert.True(_expectedCalls.Validate(_callCount), $"Unexpected number of calls to endpoint. Expected {_expectedCalls} but was {_callCount}.");
+            }
+
+            public HttpResponseMessage GetResponse()
+            {
+                if (!HasMoreCalls())
+                {
+                    throw new InvalidOperationException("This mocker limit has been reached.");
```
abdullah-cognite, line 172:
```suggestion
                    throw new InvalidOperationException("This mocker's limit has been reached.");
```

abdullah-cognite (dismissed):

abdullah-cognite (approved):

larshagencognite (approved):

larshagencognite:
risk review ok

psalaberria002:
:unicorn:

PR #244, opened 2025-03-10
Fix format warnings

lpereiracgn:


Cognite.Simulator.Utils/DefaultClasses/DefaultModelState.cs
```diff
@@ -21,8 +21,8 @@ public DefaultModelFilestate() : base()
         {
         }
 
-        public bool Processed = true;
+        private bool _processed = true;
 
-        public override bool IsExtracted => Processed;
+        public override bool IsExtracted => _processed;
```
polomani, line 26: why did this change? we need it

PR #239, opened 2025-02-20
feat: add license controller + process termination support

abdullah-cognite:
## Why is this PR needed?
This package (`dotnet-simulator-utils`) is used by the PROSPER connector to integrate the PROSPER simulator with CDF. Recently we were having issues with a customer where the PROSPER connector would not be able to run simulations (at random) because the connector was unable to acquire a license for starting the simulator, we came up with an approach where the connector would acquire a license at startup and keep holding on to it for a user defined period of time. We also noticed that occasionally when we shut down the PROSPER simulator, the license would still keep on being consumed unless the simulator process was killed.

**This PR adds** 

- `LicenseController` : To help PROSPER (or any other connector) keep track of holding the license and releasing it after a timeout
- `ProcessUtils` : To help PROSPER (or any other connector) find processes belonging to the current user and terminating them. This part also caused us to modify our Github action, since the tests for this only run on Windows. 

**This PR removes:**
- Previous process termination logic, which unfortunately didnt work as we wanted it to. Now termination is handled by each connector since there can be different error codes at which we want to terminate the process, depending upon the simulator.

PR #245, opened 2025-03-11
chore: rename build action

abdullah-cognite:
We modified the build action in this PR to use matrix: https://github.com/cognitedata/dotnet-simulator-utils/pull/239 . That caused the Github build status not to be reported correctly; since we are now reporting `build (ubuntu-latest)` and `build(windows-latest)` instead of just one `build`. This PR fixes that so now we are reporting just one `build` status.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,4 +1,4 @@
-name: Build and Test
+name: Build
 
 on:
   push:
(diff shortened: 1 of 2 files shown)


polomani (approved):
Seems to work

larshagencognite (approved):
risk review ok

PR #247, opened 2025-03-13
chore: introduce pre-commit formatting of files

lpereiracgn:
This PR introduces a pre-commit rule for formatting dotnet using the built-in tool: `dotnet format`.
The CI check for formatting is also included.

The reason for the size of this PR is the auto-generated changes.

### DIFF
--- a/.github/workflows/buildandtest.yml
+++ b/.github/workflows/buildandtest.yml
@@ -22,6 +22,8 @@ jobs:
         run: dotnet restore
       - name: Build
         run: dotnet build --no-restore
+      - name: Format
+        run: dotnet format --verify-no-changes --severity=error
       - name: Test
         run: dotnet test --no-restore --verbosity normal /p:Exclude="[*.Test]*" /p:CollectCoverage=true /p:CoverletOutputFormat=opencover /p:CoverletOutput=TestResults/ ./Cognite.Simulator.Tests/ --logger:"console"
         env:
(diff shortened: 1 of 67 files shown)


Cognite.Simulator.Tests/UtilsTests/ModelLibraryTest.cs
```diff
@@ -5,25 +5,26 @@
...
+
+using Xunit;
 
 namespace Cognite.Simulator.Tests.UtilsTests
 {
     [Collection(nameof(SequentialTestCollection))]
     public class ModelLibraryTest
     {
-        public static void CleanUp(bool deleteDirectory, StateStoreConfig stateConfig)
+        private static void CleanUp(bool deleteDirectory, StateStoreConfig stateConfig)
```
abdullah-cognite, line 27: 👍

.github/workflows/buildandtest.yml
```diff
@@ -22,6 +22,8 @@ jobs:
         run: dotnet restore
       - name: Build
         run: dotnet build --no-restore
+      - name: Format
+        run: dotnet format --verify-no-changes --severity=error
```
abdullah-cognite, line 26: Optional / non-blocking : should we do some sort of version pinning for the `dotnet format` version and mention that in the README too?

abdullah-cognite (approved):

.github/workflows/buildandtest.yml
polomani: dotnet format is a part of dotnet (starting from 6). that is why we don't even install it in the CI. so there is no need to lock it, because we already lock the dotnet version

psalaberria002 (approved):
:unicorn:

PR #252, opened 2025-04-02
docs: fix COM connector example

polomani:
this fixes the tutorial (example COM connector) to use rows, columns as input fields instead of address (excel doesn't call it address)

this was probably copied earlier from PETEX connector by mistake


abdullah-cognite (approved):

larshagencognite (approved):
🦄

PR #254, opened 2025-04-03
fix: don't call extraction pipeline when disabled

polomani:
unfortunately, as a part of the recent change, we introduced a regression by removing an extra check here: https://github.com/cognitedata/dotnet-simulator-utils/commit/800ef10ac98c6f3cd2acd38b95d3d69dfbe2ab94#r154803424
not we get null pointer exception on runtime due to this.

this reverts that change, so now both codepaths work:
1. if ext pipeline is disabled it would never be setup
2. if it is enabled but couldn't initiate on the first try, it will be retried on the next call (that was what the prev PR was about)

added a test to catch this


abdullah-cognite (approved):

larshagencognite (approved):
🦄

PR #258, opened 2025-05-06
feat: relax max model file size limit to 32GB

polomani:
We are going to enforce the file size limit on the API instead https://github.com/cognitedata/infrastructure/pull/20669 so it'll become easier for us to change these limits based on customer requirements.


abdullah-cognite (dismissed):

abdullah-cognite (approved):

larshagencognite (approved):
🦄

PR #261, opened 2025-06-05
Release v1.0.0-beta-019

polomani:
This releases a few patches to the feature that is not used in production yet.
We are working on connector support for it, so it is not (easily) possible to use until the connector with this feature is released.
The breaking change mentioned is affecting our team at most.


sagar-thalwar (approved):
👍

larshagencognite (approved):
🦄

PR #262, opened 2025-06-11
refactor model library

lpereiracgn:


Cognite.Simulator.Utils/ModelLibraryBase.cs
```diff
@@ -234,6 +228,27 @@ private async Task<T> TryReadRemoteModelRevision(string modelRevisionExternalId,
...
+
+            var model = modelVersions.FirstOrDefault();
+
+            if (model != null)
+            {
+                return model;
+            }
+
+            return null;
+        }
```
polomani, line 250: you should use `GetModelRevisionInState` instead. why is this needed?

Cognite.Simulator.Utils/ModelLibraryBase.cs
```diff
@@ -649,6 +655,67 @@ private async Task<bool> DownloadFileAsync(T modelState, bool isTemporary = fals
...
+
+                return null;
+            }
+
+            if (!shouldProcessModelRevision)
+            {
+                return await tcs.Task.ConfigureAwait(false);
+            }
+
+            return null;
```
polomani, line 716: needs some love. there is an `if (shouldProcessModelRevision)` then there is an `if (!shouldProcessModelRevision)`
1. return null; will never be reached
2. `if (!shouldProcessModelRevision)` seems redundant

Cognite.Simulator.Utils/ModelLibraryBase.cs
```diff
@@ -649,6 +655,67 @@ private async Task<bool> DownloadFileAsync(T modelState, bool isTemporary = fals
...
+
+            lock (tcs)
+            {
+                if (tcs.Task.Status == TaskStatus.WaitingForActivation)
+                {
+                    shouldProcessModelRevision = true;
+                }
+            }
+
+            if (shouldProcessModelRevision)
```
polomani, line 684: interesting, is the AI suggested? or found somewhere?

lpereiracgn: This is what I tried before with the scoping the Task to the download of the files, the example was in the RFC.

https://cognitedata.atlassian.net/wiki/spaces/PD/pages/5202149378/PO-17+Model+Library+Refactor

Cognite.Simulator.Utils/FileStorageClient.cs
```diff
@@ -70,7 +69,7 @@ public async Task<bool> DownloadFileAsync(Uri uri, string filePath, bool failLar
                     return false;
                 }
 
-                if (failLargeFiles && response.Content.Headers.ContentLength > LargeFileSize)
+                if (response.Content.Headers.ContentLength > LargeFileSize)
```
polomani, line 72: this if check should not be here

Cognite.Simulator.Utils/ModelLibraryBase.cs
```diff
@@ -668,25 +699,6 @@ private async Task SearchAndDownloadFiles(CancellationToken token)
                 await FindModelRevisions(true, token)
                     .ConfigureAwait(false);
 
-                // Find the files that are not yet saved locally (no file path)
-                var files = _state.Values
-                    .Where(f => string.IsNullOrEmpty(f.FilePath) && f.DownloadAttempts < _config.MaxDownloadAttempts)
-                    .OrderBy(f => f.UpdatedTime)
```
polomani, line 674: `.DownloadAttempts < _config.MaxDownloadAttempts`
this check doesn't happen anywhere anymore. is it on purpose?

Cognite.Simulator.Utils/ModelLibraryBase.cs
```diff
@@ -758,12 +770,8 @@ public interface IModelProvider<A, T>
...
         /// <param name="modelRevisionExternalId">Model revision external id</param>
+        /// <param name="skipCheck">If true, skips the check for filtering model revisions with processed status and filepath.</param>
         /// <returns>State object</returns>
-        Task<T> GetModelRevision(string modelRevisionExternalId);
-
-        /// <summary>
-        /// Delete all temporary model files stored in memory and on disk
-        /// </summary>
-        void WipeTemporaryModelFiles();
+        Task<T> GetModelRevision(string modelRevisionExternalId, bool skipCheck = false);
```
polomani, line 775: `bool skipCheck = false` is not easy to undestand for the consumer of the method.
name it `includeUnextracted` or something that self-explains how the behaviour of the method changes, so the user doesn't need to read the docs/source.
change the doc string accordinly

Cognite.Simulator.Utils/ModelLibraryBase.cs
```diff
@@ -649,6 +613,73 @@ private async Task<bool> DownloadFileAsync(T modelState, bool isTemporary = fals
...
+                {
+                    _logger.LogError("Error processing model revision {ExternalId}: {Message}", modelState.ExternalId, e.Message);
+                }
+
+                return null;
+            }
+            else
+            {
+                return await tcs.Task.ConfigureAwait(false);
+            }
```
abdullah-cognite, line 680: shouldn't the [SetResult](https://learn.microsoft.com/en-us/dotnet/api/system.threading.tasks.taskcompletionsource-1?view=net-9.0) method also be called on the TaskCompletionSource when we are done processing?

Cognite.Simulator.Utils/ModelLibraryBase.cs
```diff
@@ -649,6 +613,73 @@ private async Task<bool> DownloadFileAsync(T modelState, bool isTemporary = fals
...
+        /// <param name="remote">True if the model revision is remote and should be downloaded from CDF</param>
+        /// <param name="modelRevision">Model revision to process.</param>
+        private async Task<T> ProcessModelRevision(
+            T modelState,
+            CancellationToken token,
+            bool remote = false,
+            SimulatorModelRevision modelRevision = null
+        )
+        {
+            var tcs = _revisionsTasks.GetOrAdd(modelRevision.ExternalId, _ => new TaskCompletionSource<T>());
```
abdullah-cognite, line 631: when do tasks get removed from this ?

polomani:
in favour of github.com/cognitedata/dotnet-simulator-utils/pull/274

PR #274, opened 2025-07-11
fix(model-library): robustify and refactor

polomani:
Background:
Simulator connector needs models to run the simulations on. Each model is stored on the local disk and every time user uploads a new version, it gets pulled by the connector and opened in the simulator for parsing / verification.
Also, if the user triggers the simulation, the model needs to be available beforehand. This may happen either before the model has been processed, after, or even in the middle.

Why need the change:

Currently there are a number of things that arent ideal with the model file library:
1. When multiple threads access the same model revision, we download a file again a put it into a temporary folder. This adds more work to the connector, especially when things start to pile up.
2. Accesing model revision by it's externalId has major flow when user deletes and reapploads the model with the same externalId. While the code might handle some if this situations, some edge case might arise due to the code complexity.
3. Convoluted and hard to debug logic. People reporting files missing, etc.
4. We have a feature requests and with the current state of the library it is hard to make such changes.

Changes:

1. Remove the temporary state, if the model revision is being processed by the connector while other thread tries to access it we simply return the same singleton task for both threads.
2. Always compare modelRevision.externalId with the remote, if they are different -> process from scratch
3. Process all the models on the startup. Download a missing file on a fly when simulation runs happens and the file is not available.
4. Removed the "fail large file" logic, this was needed only for the files that were put into temporary state.
5. Added extensive test that checks for the correct multi-thread access to the model.

Based on top of: https://github.com/cognitedata/dotnet-simulator-utils/pull/279

NOTE: I apologise for numerous changes, but this code is so convoluted that it's impossible to achieve something by changing a single thing.


Cognite.Simulator.Tests/UtilsTests/ModelLibraryConcurrencyTest.cs
```diff
@@ -0,0 +1,175 @@
+using System;
```
polomani, line 1: move to the unit folder

Cognite.Simulator.Utils/ModelLibraryBase.cs
```diff
@@ -234,23 +224,96 @@ private async Task<T> TryReadRemoteModelRevision(string modelRevisionExternalId,
...
+                {
+                    // It will only ever be executed ONCE for a given key, by the first thread that accesses ".Value".
+                    try
+                    {
+                        return await GetOrAddModelRevisionImpl(id, remoteRevision, token).ConfigureAwait(false);
+                    }
+                    finally
+                    {
+                        _logger.LogDebug("Processing finished. Removing processing task for model revision {ExternalId}", id);
+                        _revisionsTasks.TryRemove(id, out _);
```
abdullah-cognite, line 300: I've been looking at this since the last few minutes and I think there might be an issue here. Not 100% sure though.

There's this tiny window just at the end of task completion here that can be a bit problematic. 

1: T1: Starts with externalId 1
2: T2: Starts with externalId 1
3: T1: Does its thing and reaches the remove line
4: T1: Removes the task and finishes
5: T2: Finds that there's no task with id 1, it creates a new task for the same id and starts executing it again leading to a duplicate operation.

I might be wrong, so ignore this comment if this does not make any sense.

Cognite.Simulator.Utils/ModelLibraryBase.cs
```diff
@@ -234,23 +224,96 @@ private async Task<T> TryReadRemoteModelRevision(string modelRevisionExternalId,
...
+            }
+
+            return modelState;
+        }
+
+        /// <inheritdoc/>
+        public Task<T> GetModelRevision(string modelRevisionExternalId)
+        {
+            return GetOrAddModelRevision(modelRevisionExternalId);
+        }
```
abdullah-cognite, line 279: method needs to accept and pass down the cancellation token

Cognite.Simulator.Utils/ModelLibraryBase.cs
polomani: there is a test for this. it's exaclt what this PR solves

Cognite.Simulator.Utils/ModelLibraryBase.cs
```diff
@@ -113,7 +113,6 @@ public ModelLibraryBase(
             _store = store;
             _logger = logger;
             _state = new ConcurrentDictionary<string, T>();
```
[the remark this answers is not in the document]
polomani: Not sure about this, would be great to hear someone else as well.
For now I believe having a lazy task already eliminates a range of these issues

Cognite.Simulator.Utils/ModelLibraryBase.cs
```diff
@@ -234,23 +223,101 @@ private async Task<T> TryReadRemoteModelRevision(string modelRevisionExternalId,
...
+                    try
+                    {
+                        return await GetOrAddModelRevisionImpl(id, token).ConfigureAwait(false);
+                    }
+                    finally
+                    {
+                        _logger.LogDebug("Processing finished. Removing processing task for model revision {ExternalId}", id);
+                        _revisionsTasks.TryRemove(id, out _);
+                    }
+                })
```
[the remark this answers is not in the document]
polomani: This is only true for the subsequent calls when `_revisionsTasks[id]` is still in the dictionary and it's fine. We have retries on different levels SDK / simulator, etc.

Cognite.Simulator.Utils/ModelLibraryBase.cs
```diff
@@ -50,11 +50,11 @@ public abstract class ModelLibraryBase<A, T, U, V> : IModelProvider<A, T>
...
         /// <summary>
-        /// Temporary model states that are used once and then deleted after the run.
-        /// We keep them in memory to avoid saving them to the state store and hence to avoid the multi-threading issues.
-        /// This state is used to store temporary model file paths, used only in case a run item is received before the model file has been downloaded.
+        /// Dictionary holding tasks for processing model revisions. The keys are the external ids of the
+        /// CDF simulator model revisions and the values are LazyTask objects for the state objects of type <typeparamref name="T"/>.
+        /// This ensures that only one thread processes a given model revision at a time.
         /// </summary>
-        public Dictionary<string, T> _temporaryState { get; private set; }
+        private readonly ConcurrentDictionary<string, Lazy<Task<T>>> _revisionsTasks = new();
```
abdullah-cognite, line 57: async version of the conncurrent dictionary

Cognite.Simulator.Utils/ModelLibraryBase.cs
```diff
@@ -745,6 +723,14 @@ await _store.StoreExtractionState(
             token).ConfigureAwait(false);
         }
 
+        /// <summary>
+        /// Disposes the model library and releases any resources it holds.
+        /// </summary>
+        public void Dispose()
+        {
+            _revisionsTasks?.Dispose();
+            GC.SuppressFinalize(this);
```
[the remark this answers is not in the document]
polomani: @gemini-code-assist when I remove it the linter says: Change ModelLibraryBase<A, T, U, V>.Dispose() to call GC.SuppressFinalize(object). This will prevent derived types that introduce a finalizer from needing to re-implement 'IDisposable' to call it.[CA1816](https://learn.microsoft.com/dotnet/fundamentals/code-analysis/quality-rules/ca1816)

Cognite.Simulator.Utils/ModelLibraryBase.cs
polomani: Made this more sequencial now, so the worst case is that next call (if unfortunate) will call `GetOrAddModelRevisionImpl`, which will exit after `TryReadRemoteModelRevision` if no work needed (nothing got updated)

abdullah-cognite (dismissed):

Cognite.Simulator.Utils/ModelLibraryBase.cs
```diff
@@ -234,23 +231,68 @@ private async Task<T> TryReadRemoteModelRevision(string modelRevisionExternalId,
             return null;
         }
 
-        /// <inheritdoc/>
-        public async Task<T> GetModelRevision(string modelRevisionExternalId)
+        /// <summary>
+        /// If model revision is not found in the local state, this method will try to read it from CDF, download the file and extract the model information.
+        /// If the model revision is already in the local state and has not been changed, it will return the existing state.
```
einarmo, line 236: This doc comment isn't true, I think? You unconditionally call `TryReadRemoteModelRevision` on the first line, which will make a call to CDF.

Wouldn't it be better to cache by external ID, and look that up first?

Cognite.Simulator.Utils/ModelLibraryBase.cs
```diff
@@ -234,23 +231,68 @@ private async Task<T> TryReadRemoteModelRevision(string modelRevisionExternalId,
...
+            var modelRevision = await TryReadRemoteModelRevision(modelRevisionExternalId, token).ConfigureAwait(false);
+
+            if (modelRevision == null)
+            {
+                _logger.LogError("Model revision {ModelRevisionExternalId} not found in CDF", modelRevisionExternalId);
+                return null;
+            }
 
-            var model = modelVersions.FirstOrDefault();
+            _state.TryGetValue(modelRevision.Id.ToString(), out var modelState);
```
einarmo, line 250: Why the `Id.ToString()`? This smells a little to me.

Cognite.Simulator.Utils/ModelStateBase.cs
```diff
@@ -47,6 +47,17 @@ public virtual bool ShouldProcess()
             return !string.IsNullOrEmpty(FilePath) && CanRead && !isParsedBefore;
         }
 
+        /// <summary>
+        /// Indicates if the model file has been downloaded and the file exists on the disk.
+        /// </summary>
+        public bool Downloaded
```
einarmo, line 53: To me, this implementation is a little scary. I don't know exactly where this value is checked, but what happens if a file is in the _process_ of being downloaded? Won't this return `true` in that case? Potentially incorrectly informing consumers that the file is ready?

einarmo (commented):
Overall good changes, but I have some questions

Cognite.Simulator.Utils/ModelStateBase.cs
polomani: nope. we only set filepath after something has been downloaded, not before / in the process

Cognite.Simulator.Utils/ModelLibraryBase.cs
polomani: I agree this can be improved, but it's been stored like this (key as strings) even before this PR, so I don't want to pull more changes in (I can create a ticket ,maybe?)

Cognite.Simulator.Utils/ModelLibraryBase.cs
polomani: fixed the wording

abdullah-cognite (approved):

einarmo:
:unicorn:

PR #288, opened 2025-08-13
chore: release v1.0.0-beta-020

polomani:
All the changes mentioned in the changelog.
This is intermediate release, we will be releasing more changes on top of it. We are the main (and only) user of the library, so this release is needed to unblock ourselves, while we are still working on some polishing and performance improvements.


abdullah-cognite (approved):

polomani:
🦄

PR #291, opened 2025-08-14
feat(modellib): support more than 1000 files per model

polomani:
One model can consist of 1000+ files. Currently we download them one by one to not stress the API and this PR doesn't change the behaviour.

One limitation that we have is that `files.RetrieveAsync` doesn't support getting metadata of more than 1000 files at once.
This is a safe operation to do concurrently, so I am adding an extension method to support chunking.

I replaced the method call, so the lines are covered by the previous tests, though I added one more for a happy path as well.

### DIFF
--- a/Cognite.Simulator.Extensions/FilesExtensions.cs
+++ b/Cognite.Simulator.Extensions/FilesExtensions.cs
@@ -23,7 +23,7 @@ public static class FilesExtensions
 
     /// <summary>
     /// Retrieves file medatadata by their IDs in chunks, ignores unknown IDs.
-    /// Each batch can contain up to 1000 file IDs and is be processed in parallel (up to 5 batches at a time).
+    /// Each batch can contain up to 1000 file IDs and is processed in parallel to others (up to 5 batches at a time).
     /// </summary>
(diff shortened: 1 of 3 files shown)


abdullah-cognite (dismissed):

Cognite.Simulator.Extensions/FilesExtensions.cs
```diff
@@ -9,6 +18,51 @@ namespace Cognite.Simulator.Extensions;
...
+
+        var fileIdsByChunks = internalIds
+            .ChunkBy(chunkSize);
+
+        var generators = fileIdsByChunks
+            .Select<IEnumerable<long>, Func<Task>>(
+            (chunk, _) => async () =>
+            {
+                var found = await cdfFiles
+                    .RetrieveAsync(chunk, true, token)
```
msusag, line 54:
```suggestion
                    .RetrieveAsync(chunk, ignoreUnknownIds: true, token)
```

msusag:
Looks good, but added a few minor/tiny suggestions to be considered before merging.

lpereiracgn (approved):

PR #296, opened 2025-08-20
chore: release beta v1.0.0-beta-021

polomani:
This is just mainly a release, but also contains a hot fix for a prev PR: https://github.com/cognitedata/dotnet-simulator-utils/pull/284 (and the fix is taking most of the line changes due to the test)
Basically, fixes the edge case when there are 0 files to download by doing an early return. Tests are extended to cover that.

### Features

* **Improved Model Library Performance**: State is now restored from local cache on startup, avoiding redundant model processing
* **File Download Deduplication**: Models sharing the same files no longer re-download duplicates, significantly improving performance for multi-file models

### Other changes

* Removed unused properties from model state for cleaner codebase


abdullah-cognite (dismissed):

abdullah-cognite (approved):

msusag:
🦄

PR #300, opened 2025-08-25
fix(connector-runtime): update ModelDependencies definition on startup

polomani:
When connector starts-up it normally upserts the "Simulator" definition on remote. In the recent update we added a `ModelDependencies` property, but we forgot to update them on startup. This fixes that, also moved the implementation to the `SimulatorsExtensions` class as it makes sense to keep it closer to ther similar extensions.


abdullah-cognite (dismissed):
👍

polomani:
🦄

abdullah-cognite (approved):

PR #302, opened 2025-08-27
feat(modellib): support bulk upload

polomani:
One model can consist of 1000+ files. Currently get the metadata of all the files at once, and then we download them one by one to not stress the API. 

One limitatio we have is that `files.RetrieveAsync` doesn't support getting metadata of more than 1000 files at once. Also downloading 1000 files one by one can take time.

From now on we split the request in chunks of 10 and we get metadata / download links for the whole batch at once, then we also download files for the given batch in parallel. Multiple batches are still downloaded sequencially, same as multiple models.

Tests: sort of covered by the prev implementation.


polomani:
actually, this is more complex and most likely not better in terms of speed

PR #304, opened 2025-09-18
fix(routine-library): optimise memory consumption [POFSP-1221]

polomani:
So far we've been saving all the latest version of routines in memory.
This used to work fine and was needed to read the cron params and schedule routine executions based on that. This in-memory cache was also used to run the simulation.

Going forward we might start having larger (few MBs routine revisions) or having simply much larger amount of routines in total.
So to eliminate any future out-of-memory issues with change the logic to only store the most important fields (`IncludeAllFields=false` while listing the latest versions).
Afterwards we store the relevant fields into `SimulatorRoutineRevisionInfo` abstraction.

In order to run a simulation, we'll make a direct call to the API instead of fetching the routine from memory.

Tests: I consider this being covered by the current integration test suite as it's mostly a refactor

### DIFF
--- a/Cognite.Simulator.Extensions/Types.cs
+++ b/Cognite.Simulator.Extensions/Types.cs
@@ -29,37 +29,37 @@ namespace Cognite.Simulator.Extensions
         /// <summary>
         /// The unique identifier of the routine revision.
         /// </summary>
-        public long Id { get; set; }
+        public long Id { get; private set; }
 
         /// <summary>
         /// Routine revision external id
         /// </summary>
-        public string ExternalId { get; set; }
+        public string ExternalId { get; private set; }
 
         /// <summary>
         /// Routine external id
         /// </summary>
-        public string RoutineExternalId { get; set; }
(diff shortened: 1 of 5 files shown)


abdullah-cognite (dismissed):
LGTM, I was thinking...should we periodically log memory usage as well for the connector? (GC.GetTotalMemory(false);)

Cognite.Simulator.Extensions/Types.cs
```diff
@@ -24,6 +26,11 @@ public class SimulatorModelInfo
     /// </summary>
     public class SimulatorRoutineRevisionInfo
     {
+        /// <summary>
+        /// The unique identifier of the routine revision.
+        /// </summary>
+        public long Id { get; set; }
```
msusag, line 32: Should these properties be immutable, i.e. use `{ get; init; }`

polomani: we are on .net 8, but I can make it `internal set` instead

abdullah-cognite (approved):

PR #305, opened 2025-09-23
Release v1.0.0-beta-023

abdullah-cognite:
### Features
* **Connector Startup Improvement**: Added 1-minute memory caching to prevent redundant API calls when validating locally stored model revisions during connector startup, eliminating rate limiting issues.  
* **Model Library**: Introduced concurrent chunking support for retrieving metadata from 1000+ model files.
* **Routine Library**: Optimized memory usage by storing only essential routine fields in memory and fetching complete data on-demand to prevent out-of-memory issues.
### Other changes
* **Routine Library**: Refactored routine revision item references to use long IDs directly instead of string keys.


polomani (approved):
🦄

PR #307, opened 2025-09-25
fix: reset download attempt counter on successful downloads 

abdullah-cognite:
The file read error popped up again today, where the connector is not attempting to re-download a file if it doesn't exist on the disk.

This PR attempts to address that by simply resetting the download attempt counter after a succesful download attempt, since this was preventing the connector to re-download a file even if it didnt exist .


Cognite.Simulator.Utils/ModelLibraryBase.cs
```diff
@@ -297,6 +304,8 @@ private async Task<T> GetOrAddModelRevisionImpl(string modelRevisionExternalId,
                 await ExtractModelInformationAndPersist(modelState, token).ConfigureAwait(false);
             }
 
+            modelState.DownloadAttempts = 0;
```
polomani, line 307: this should be only reset after the model is downloaded, otherwise we get an endless retry loop
I think it's best to put it inside the `DownloadAllModelFilesAsync`

before the end of the function we could do:
```
if (allFilesDownloaded) {
    modelState.DownloadAttempts = 0;
}
```

Cognite.Simulator.Utils/ModelLibraryBase.cs
```diff
@@ -709,6 +707,10 @@ private async Task<bool> DownloadAllModelFilesAsync(T modelState)
                         fileId, modelState.ExternalId);
                 }
                 allFilesDownloaded &= downloaded;
+                if (allFilesDownloaded)
+                {
+                    modelState.DownloadAttempts = 0;
+                }
```
polomani, line 713: again, wrong place.
should be after the curly brace, one line lower

polomani (approved):
🦄

PR #313, opened 2025-11-11
feat(modellib): use SimulatorExternalIds filter to list model revision

mohammad-arif662:
This PR is to add `SimulatorExternalIds` filter to model revision queries to get filtered revisions instead of getting all the revisions and then filter.


polomani (approved):
🦄

PR #328, opened 2026-02-05
fix: connector doesn't recover from failure after network issue

mohammad-arif662:
## Problem
When a simulation was running and the system lost internet connectivity, the connector would terminate unexpectedly, leaving the run stuck in "running" status indefinitely. It happens because `PublishConnectorStatus` was declared as `async void`, which cannot be awaited by the caller and causes unhandled exception type `HttpRequestException` to crash the process.

## Fix
Changed `PublishConnectorStatus` from `async void` to `async Task` and added proper await calls. Now network exceptions propagate to the outer try/catch block which allows the connector to continue running and recover when connectivity is restored and marks the run as "failure".


Cognite.Simulator.Utils/SimulationRunnerBase.cs
```diff
@@ -247,7 +247,7 @@ await InitSimulationRun(
                             if (!skipped)
                             {
                                 _logger.LogDebug("Simulation run finished for run {Id}", runId);
-                                PublishConnectorStatus(ConnectorStatus.IDLE, token);
+                                await PublishConnectorStatusAsync(ConnectorStatus.IDLE, token).ConfigureAwait(false);
```
abdullah-cognite: can we add a test with mocks where we fail this call ?

abdullah-cognite (changes requested):

Cognite.Simulator.Tests/UtilsTests/TestUtilities.cs
```diff
@@ -275,6 +275,40 @@ public static HttpResponseMessage MockSimulatorsIntegrationsEndpoint()
...
+                ""createdTime"": {now},
+                ""lastUpdatedTime"": {now}
+            }}";
+            return OkItemsResponse(item);
+        }
+
+        public static HttpResponseMessage MockSimulationRunsListEmptyEndpoint()
+        {
+            return OkItemsResponse("");
+        }
```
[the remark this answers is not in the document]
mohammad-arif662: added it

Cognite.Simulator.Tests/UtilsTests/SimulationRunnerUnitTest.cs
```diff
@@ -0,0 +1,148 @@
...
+            }
+            catch (OperationCanceledException) { }
+
+            VerifyLog(mockedSimulationRunnerLogger, LogLevel.Warning, "Simulation run 12345 failed with error:", Times.AtLeastOnce(), true);
+            VerifyLog(mockedSimulationRunnerLogger, LogLevel.Warning, "Network error during callback update", Times.AtLeastOnce(), true);
+            VerifyLog(mockedConnectorLogger, LogLevel.Information, "Connector started", Times.AtLeast(2), true);
+        }
+
+
+        private class EmptySimulatorAutomationClient :
```
abdullah-cognite: this one is nearly identical to the one in `ConnectorRuntimeUnitTest.cs` . I would recommend to extract a shared base.

Cognite.Simulator.Tests/UtilsTests/SimulationRunnerUnitTest.cs
```diff
@@ -0,0 +1,148 @@
...
+            DefaultConnectorRuntime<DefaultAutomationConfig, DefaultModelFilestate, DefaultModelFileStatePoco>.ConnectorName = "Empty";
+            DefaultConnectorRuntime<DefaultAutomationConfig, DefaultModelFilestate, DefaultModelFileStatePoco>.SimulatorDefinition = SeedData.SimulatorCreate;
+
+            try
+            {
+                await DefaultConnectorRuntime<DefaultAutomationConfig, DefaultModelFilestate, DefaultModelFileStatePoco>.Run(mockedLogger.Object, cts.Token);
+            }
+            catch (OperationCanceledException) { }
+
+            VerifyLog(mockedSimulationRunnerLogger, LogLevel.Warning, "Simulation run 12345 failed with error:", Times.AtLeastOnce(), true);
```
abdullah-cognite: extract the 12345 to a common constant in the mock and use that here

Cognite.Simulator.Tests/UtilsTests/SimulationRunnerUnitTest.cs
```diff
@@ -0,0 +1,148 @@
...
+            };
+            DefaultConnectorRuntime<DefaultAutomationConfig, DefaultModelFilestate, DefaultModelFileStatePoco>.ConnectorName = "Empty";
+            DefaultConnectorRuntime<DefaultAutomationConfig, DefaultModelFilestate, DefaultModelFileStatePoco>.SimulatorDefinition = SeedData.SimulatorCreate;
+
+            try
+            {
+                await DefaultConnectorRuntime<DefaultAutomationConfig, DefaultModelFilestate, DefaultModelFileStatePoco>.Run(mockedLogger.Object, cts.Token);
+            }
+            catch (OperationCanceledException) { }
+
```
abdullah-cognite: also add a VerifyLog for the starting simulation run line

abdullah-cognite (changes requested):

Cognite.Simulator.Tests/UtilsTests/SimulationRunnerUnitTest.cs
```diff
@@ -0,0 +1,102 @@
...
+                services.AddSingleton(mockedConnectorLogger.Object);
+            };
+            DefaultConnectorRuntime<DefaultAutomationConfig, DefaultModelFilestate, DefaultModelFileStatePoco>.ConnectorName = "Empty";
+            DefaultConnectorRuntime<DefaultAutomationConfig, DefaultModelFilestate, DefaultModelFileStatePoco>.SimulatorDefinition = SeedData.SimulatorCreate;
+
+            try
+            {
+                await DefaultConnectorRuntime<DefaultAutomationConfig, DefaultModelFilestate, DefaultModelFileStatePoco>.Run(mockedLogger.Object, cts.Token);
+            }
+            catch (OperationCanceledException) { }
```
polomani: you need to check whether the connector threw `OperationCanceledException`, catching this blindly would give a false confidence, because there might be other errors that stopped the connector (like in this case, `Network error`)

Cognite.Simulator.Tests/UtilsTests/TestUtilities.cs
```diff
@@ -275,6 +275,40 @@ public static HttpResponseMessage MockSimulatorsIntegrationsEndpoint()
...
+        public static HttpResponseMessage MockSimulatorIntegrationsListEndpoint()
+        {
+            var item = $@"{{
+                ""id"": 999,
+                ""externalId"": ""{SeedData.TestIntegrationExternalId}"",
+                ""simulatorExternalId"": ""{SeedData.TestSimulatorExternalId}"",
+                ""dataSetId"": {SeedData.TestDataSetId}
+            }}";
+            return OkItemsResponse(item);
+        }
```
polomani: I think you can reuse the `MockSimulatorsIntegrationsEndpoint` if you add lacking fields to it?

Cognite.Simulator.Tests/UtilsTests/TestUtilities.cs
```diff
@@ -286,6 +320,25 @@ public static HttpResponseMessage MockSimulatorRoutineRevEndpoint()
...
+                ""modelExternalId"": ""test-model"",
+                ""name"": ""Test routine revision"",
+                ""dataSetId"": 123,
+                ""createdTime"": {now},
+                ""lastUpdatedTime"": {now},
+                ""configuration"": {{}}
+            }}";
+            return OkItemsResponse(item);
+        }
+
```
polomani: you can just reuse `MockSimulatorRoutineRevEndpoint` if you add `simulatorIntegrationExternalId` to it

Cognite.Simulator.Tests/UtilsTests/TestUtilities.cs
```diff
@@ -307,6 +360,21 @@ public static HttpResponseMessage MockSimulatorModelRevEndpoint()
             return OkItemsResponse(item);
         }
 
+        public static HttpResponseMessage MockSimulatorModelRevListEndpoint()
```
polomani: `MockSimulatorModelRevEndpoint` should just work as good as this one

Cognite.Simulator.Tests/UtilsTests/TestUtilities.cs
mohammad-arif662: reused it

Cognite.Simulator.Tests/UtilsTests/TestUtilities.cs
mohammad-arif662: reused it

Cognite.Simulator.Tests/UtilsTests/TestUtilities.cs
mohammad-arif662: reused it

abdullah-cognite (dismissed):
looks good

polomani (approved):
🦄

PR #333, opened 2026-04-22
feat: fetch the simulation runs from /poll endpoint when config is enabled

mohammad-arif662:
PR is to enable dynamic distribution of simulation runs across multiple connectors via a new poll endpoint.

- Add `LoadBalancerEnabled` flag (default: false) and `SimulationRunPollLimit` (default 10) to `ConnectorConfig`
- When enabled, use new `PollSimulationRunsAsync` endpoint to claim ready runs instead of integration ID filtering
- Included connector externalId in run status callbacks when load balancer is active
- Bump `CogniteSdk` to `5.7.0` to support poll endpoint
- Added unit tests


Cognite.Simulator.Tests/UtilsTests/SimulationRunnerUnitTest.cs
```diff
@@ -95,5 +95,105 @@ public async Task TestSimulationRunner_CallbackNetworkError_HandledGracefully()
...
+            }
+
+            Assert.True(
+                thrownException == null || thrownException is OperationCanceledException,
+                $"Expected OperationCanceledException or no exception, but got: {thrownException?.GetType().Name}: {thrownException?.Message}");
+
+            VerifyLoadBalancerEndpointUsage(mockedSimulationRunnerLogger, loadBalancerEnabled);
+        }
+
+        private List<SimpleRequestMocker> BuildNetworkMocksForLoadBalancer(bool loadBalancerEnabled)
```
vikramlc-cognite: Can you check if we can use this method in this test: TestSimulationRunner_CallbackNetworkError_HandledGracefully as well ?

Cognite.Simulator.Utils/SimulationRunnerBase.cs
```diff
@@ -109,14 +111,21 @@ private async Task<SimulationRun> UpdateSimulationRunStatus(
...
+
             var res = await _cdfSimulators.SimulationRunCallbackAsync(
-                new SimulationRunCallbackItem()
-                {
-                    Id = runId,
-                    Status = status,
-                    StatusMessage = message,
-                    SimulationTime = simulationTime
-                }, token).ConfigureAwait(false);
+                callbackItem, token).ConfigureAwait(false);
```
vikramlc-cognite, line 122: Add a test for callback called with SimulatorIntegrationExternalId.

Cognite.Simulator.Utils/SimulationRunnerBase.cs
```diff
@@ -127,6 +136,23 @@ private async Task<IEnumerable<SimulationRun>> FindSimulationRunsWithStatus(
...
+                        new SimulationRunPollItem()
+                        {
+                            SimulatorIntegrationExternalId = connectorName,
+                            Limit = DefaultSimulationRunsPollLimit
+                        }
+                    }, token).ConfigureAwait(false);
+                return pollResult.Items;
+            }
+
+            _logger.LogDebug("Fetching simulation runs via list endpoint with integration filter");
```
vikramlc-cognite: I don't think these debug loggers are needed.

Cognite.Simulator.Utils/SimulationRunnerBase.cs
```diff
@@ -127,6 +136,23 @@ private async Task<IEnumerable<SimulationRun>> FindSimulationRunsWithStatus(
         {
             var connectorName = _connectorConfig.GetConnectorName();
 
+            // Poll endpoint is only for ready status. Stale run (running status) always uses list endpoint.
+            if (_connectorConfig.LoadBalancerEnabled && status == SimulationRunStatus.ready)
```
vikramlc-cognite: Question: Shouldn't status == SimulationRunStatus.queued instead? Status of run moves to ready only when the assignment happens in the poll endpoint correct?

Cognite.Simulator.Utils/SimulationRunnerBase.cs
mohammad-arif662: This is covered when `[InlineData(true)]` in test `TestSimulationRunner_FetchEndpointSelection`

Cognite.Simulator.Utils/Configuration.cs
```diff
@@ -215,6 +215,12 @@ public class ConnectorConfig
         /// </summary>
         public LicenseCheckConfig LicenseCheck { get; set; } = new LicenseCheckConfig();
 
+        /// <summary>
+        /// When true, the connector claims runs via poll endpoint
+        /// instead of the legacy static SimulatorIntegrationExternalId filter.
+        /// </summary>
+        public bool LoadBalancerEnabled { get; set; }
```
polomani:
```suggestion
        public bool SimulationRunLoadBalancerEnabled { get; set; }
```
I think better to make it obvious

Cognite.Simulator.Utils/SimulationRunnerBase.cs
```diff
@@ -109,14 +111,21 @@ private async Task<SimulationRun> UpdateSimulationRunStatus(
...
+                Id = runId,
+                Status = status,
+                StatusMessage = message,
+                SimulationTime = simulationTime
+            };
+
+            if (_connectorConfig.LoadBalancerEnabled)
+            {
+                callbackItem.SimulatorIntegrationExternalId = _connectorConfig.GetConnectorName();
+            }
```
polomani:
```suggestion
            var callbackItem = new SimulationRunCallbackItem()
            {
                Id = runId,
                Status = status,
                StatusMessage = message,
                SimulationTime = simulationTime,
                SimulatorIntegrationExternalId = _connectorConfig.LoadBalancerEnabled ? _connectorConfig.GetConnectorName() : null
            };
```

Cognite.Simulator.Utils/SimulationRunnerBase.cs
polomani: seems right the way it is, `ready` is what we get back

Cognite.Simulator.Utils/SimulationRunnerBase.cs
```diff
@@ -127,6 +136,23 @@ private async Task<IEnumerable<SimulationRun>> FindSimulationRunsWithStatus(
...
+            if (_connectorConfig.LoadBalancerEnabled && status == SimulationRunStatus.ready)
+            {
+                _logger.LogDebug("Fetching simulation runs via poll endpoint");
+                var pollResult = await _cdfSimulators.PollSimulationRunsAsync(
+                    new[]
+                    {
+                        new SimulationRunPollItem()
+                        {
+                            SimulatorIntegrationExternalId = connectorName,
+                            Limit = DefaultSimulationRunsPollLimit
```
polomani: make this parametrized from ConnectorConfig (can be with default value, that is ok)

Cognite.Simulator.Utils/SimulationRunnerBase.cs
```diff
@@ -127,6 +136,23 @@ private async Task<IEnumerable<SimulationRun>> FindSimulationRunsWithStatus(
         {
             var connectorName = _connectorConfig.GetConnectorName();
 
+            // Poll endpoint is only for ready status. Stale run (running status) always uses list endpoint.
```
polomani: > Stale run (running status) always uses list endpoint.

what does this comment mean? can you rephrase?

Cognite.Simulator.Tests/UtilsTests/SimulationRunnerUnitTest.cs
```diff
@@ -95,5 +95,105 @@ public async Task TestSimulationRunner_CallbackNetworkError_HandledGracefully()
...
+            }
+            else
+            {
+                commonMocks.Add(new SimpleRequestMocker(uri => uri.Contains("/simulators/runs/list"), MockSimulationRunsListEndpoint, 1));
+                commonMocks.Add(new SimpleRequestMocker(uri => uri.Contains("/simulators/runs/list"), MockSimulationRunsListEmptyEndpoint));
+            }
+
+            commonMocks.AddRange(new[]
+            {
+                new SimpleRequestMocker(uri => uri.Contains("/simulators/run/callback"), () => OkItemsResponse($@"{{""id"": {SeedData.TestSimulationRunId}, ""status"": ""running""}}")),
```
polomani: how many calls do you expect? better to limit

Cognite.Simulator.Tests/UtilsTests/SimulationRunnerUnitTest.cs
```diff
@@ -95,5 +95,105 @@ public async Task TestSimulationRunner_CallbackNetworkError_HandledGracefully()
...
+            catch (Exception ex)
+            {
+                thrownException = ex;
+            }
+
+            Assert.True(
+                thrownException == null || thrownException is OperationCanceledException,
+                $"Expected OperationCanceledException or no exception, but got: {thrownException?.GetType().Name}: {thrownException?.Message}");
+
+            VerifyLoadBalancerEndpointUsage(mockedSimulationRunnerLogger, loadBalancerEnabled);
```
polomani: add
```
foreach (var mocker in networkMocks)
            {
                mocker.AssertCallCount();
            }
```

Cognite.Simulator.Tests/UtilsTests/SimulationRunnerUnitTest.cs
```diff
@@ -95,5 +95,105 @@ public async Task TestSimulationRunner_CallbackNetworkError_HandledGracefully()
...
+            bool loadBalancerEnabled)
+        {
+            if (loadBalancerEnabled)
+            {
+                VerifyLog(logger, LogLevel.Debug, "Fetching simulation runs via poll endpoint", Times.Exactly(1), true);
+                VerifyLog(logger, LogLevel.Debug, "Fetching simulation runs via list endpoint with integration filter", Times.Exactly(1), true);
+            }
+            else
+            {
+                VerifyLog(logger, LogLevel.Debug, "Fetching simulation runs via list endpoint with integration filter", Times.Exactly(2), true);
```
polomani: instead of asserting the logger, you can assert the http calls https://github.com/cognitedata/dotnet-simulator-utils/pull/333/changes#r3130107526

then you don't need the debug statements, perhaps

Cognite.Simulator.Tests/UtilsTests/TestUtilities.cs
```diff
@@ -300,6 +301,24 @@ public static HttpResponseMessage MockSimulationRunsListEmptyEndpoint()
...
+                ""simulatorIntegrationExternalId"": ""{SeedData.TestIntegrationExternalId}"",
+                ""routineRevisionExternalId"": ""test-routine-rev"",
+                ""modelRevisionExternalId"": ""test-model-rev"",
+                ""routineExternalId"": ""test-routine"",
+                ""runType"": ""external"",
+                ""createdTime"": {now},
+                ""lastUpdatedTime"": {now}
+            }}";
+            return OkItemsResponse(item);
+        }
```
polomani: is it the same as MockSimulationRunsListEndpoint? can you reuse?

Cognite.Simulator.Tests/UtilsTests/SimulationRunnerUnitTest.cs
mohammad-arif662: I think better to keep it separate because `BuildNetworkMocksForLoadBalancer` has load balancer mock based on enable disable

Cognite.Simulator.Utils/SimulationRunnerBase.cs
mohammad-arif662: added config in ConnectorConfig with default

Cognite.Simulator.Tests/UtilsTests/SimulationRunnerUnitTest.cs
mohammad-arif662: added limits in mocks

Cognite.Simulator.Tests/UtilsTests/TestUtilities.cs
mohammad-arif662: reused `MockSimulationRunsListEndpoint`

Cognite.Simulator.Utils/SimulationRunnerBase.cs
vikramlc-cognite: I was referring to adding an assert on if the callback API is called with simulatorIntegrationExternalId as null or with some value. Can you add that as part of the above test you mentioned: **TestSimulationRunner_FetchEndpointSelection** ?

mohammad-arif662: we are already asserting the /callback endpoint in the test. To validate request body we have to change test utilities `SimpleRequestMocker` and `MockRequestsAsync` which might not be good idea

Cognite.Simulator.Tests/UtilsTests/SimulationRunnerUnitTest.cs
vikramlc-cognite: There is code duplication here as we are creating same large object twice which can be avoided if we create a common method which can be called **BuildNetworkMocks** with maybe few params based on load balancer or not and get the object you require.
Would recommend to add **BuildNetworkMocks** in **TestUtilities.cs**.

Cognite.Simulator.Utils/SimulationRunnerBase.cs
vikramlc-cognite: Changes are very simple and minor.
You need to create a callback mocker like below in BuildNetworkMocksForLoadBalancer and return it as part of the response.
Then in your test, use the callback mocker to assert as below. Can you check if this works?
```
if (simulationRunLoadBalancerEnabled)
{
    callbackMocker.AssertAllRequestBodiesContain(SeedData.TestIntegrationExternalId);
}
else
{
    callbackMocker.AssertNoRequestBodyContains(SeedData.TestIntegrationExternalId);
}
```

vikramlc-cognite: Non-blocker for me for this PR.

Cognite.Simulator.Tests/UtilsTests/SimulationRunnerUnitTest.cs
vikramlc-cognite: Non blocker for me for this PR.

vikramlc-cognite (dismissed):

Cognite.Simulator.Utils/Configuration.cs
```diff
@@ -215,6 +215,17 @@ public class ConnectorConfig
...
+        /// When true, the connector claims runs via poll endpoint
+        /// instead of the legacy static SimulatorIntegrationExternalId filter.
+        /// </summary>
+        public bool SimulationRunLoadBalancerEnabled { get; set; }
+
+        /// <summary>
+        /// Maximum number of simulation runs to claim per poll request when simulation run load balancer is enabled.
+        /// </summary>
+        public int SimulationRunsPollLimit { get; set; } = 10;
+
```
polomani, line 228: one is plural `runs` and one is singular `run`, you have to align them, I think we have to go with singular because of `SimulationRunTolerance`.

though `FetchRunsInterval` is plural, omg..

version
```diff
@@ -1 +1 @@
-1.0.0-beta-026
+1.0.0-beta-027
```
polomani, line 1: where is changelog?

Cognite.Simulator.Utils/SimulationRunnerBase.cs
```diff
@@ -127,6 +130,21 @@ private async Task<IEnumerable<SimulationRun>> FindSimulationRunsWithStatus(
         {
             var connectorName = _connectorConfig.GetConnectorName();
 
+            // When simulation run load balancer is enabled, poll is used for ready status and list endpoint for running status.
+            if (_connectorConfig.SimulationRunLoadBalancerEnabled && status == SimulationRunStatus.ready)
+            {
+                var pollResult = await _cdfSimulators.PollSimulationRunsAsync(
+                    new[]
```
polomani: nit:

Image

version
polomani: see https://github.com/cognitedata/dotnet-simulator-utils/blob/edc51a3a54e22c92aca99b032331a868f6d65cee/CHANGELOG.md?plain=1#L134

Cognite.Simulator.Utils/Configuration.cs
mohammad-arif662: made it singular

version
mohammad-arif662: added changelog

polomani:
🦄

vikramlc-cognite (approved):

PR #335, opened 2026-04-30
feat(simint): Fetch both ready and queued runs when load balancer is enabled

mohammad-arif662:
When the simulation run load balancer is enabled, fetch ready runs from the `/list` endpoint first, then fetch queued runs from `/poll` with an adjusted limit. This prevents duplicate runs and ensures we never exceed the configured poll limit.

- Updated `FindSimulationRunsWithStatus` to call `/list` before `/poll` when load balancer is enabled, adjusting the poll limit by the number of ready runs returned
- Updated `TestSimulationRunner_FetchEndpointSelection` to verify both endpoints are called in the correct order


Cognite.Simulator.Utils/SimulationRunnerBase.cs
```diff
@@ -153,6 +139,33 @@ private async Task<IEnumerable<SimulationRun>> FindSimulationRunsWithStatus(
...
             };
+
+            // When simulation run load balancer is enabled, for ready status fetch both ready and queued runs
+            if (_connectorConfig.SimulationRunLoadBalancerEnabled && status == SimulationRunStatus.ready)
+            {
+                var listResult = await _cdfSimulators
+                    .ListSimulationRunsAsync(query, token)
+                    .ConfigureAwait(false);
+
+                var readyRuns = listResult.Items.ToList();
```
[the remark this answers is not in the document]
mohammad-arif662: sdk responses guarantee that Items is never null, it's an empty collection if there are no items

Cognite.Simulator.Utils/SimulationRunnerBase.cs
```diff
@@ -153,6 +139,33 @@ private async Task<IEnumerable<SimulationRun>> FindSimulationRunsWithStatus(
...
+                {
+                    var pollResult = await _cdfSimulators.PollSimulationRunsAsync(
+                        [
+                            new SimulationRunPollItem()
+                            {
+                                SimulatorIntegrationExternalId = connectorName,
+                                Limit = remainingLimit
+                            }
+                        ], token).ConfigureAwait(false);
+                    readyRuns.AddRange(pollResult.Items);
```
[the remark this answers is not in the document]
mohammad-arif662: same as above

Cognite.Simulator.Utils/SimulationRunnerBase.cs
```diff
@@ -153,6 +139,33 @@ private async Task<IEnumerable<SimulationRun>> FindSimulationRunsWithStatus(
...
                 }
             };
+
+            // When simulation run load balancer is enabled, for ready status fetch both ready and queued runs
+            if (_connectorConfig.SimulationRunLoadBalancerEnabled && status == SimulationRunStatus.ready)
+            {
+                var listResult = await _cdfSimulators
+                    .ListSimulationRunsAsync(query, token)
+                    .ConfigureAwait(false);
+
```
polomani: this line is the same as `var runsResult` lower. you need to move it up and reuse

polomani: or delete, and move the block down and put after
`var runsResult = await _cdfSimulators`

polomani (changes requested):
needs a bit of refactor

Cognite.Simulator.Utils/SimulationRunnerBase.cs
```diff
@@ -153,11 +139,34 @@ private async Task<IEnumerable<SimulationRun>> FindSimulationRunsWithStatus(
...
+            var listResult = await _cdfSimulators
                 .ListSimulationRunsAsync(query, token)
                 .ConfigureAwait(false);
 
-            return runsResult.Items;
+            // When simulation run load balancer is enabled, for ready status fetch queued runs up to SimulationRunPollLimit
+            if (_connectorConfig.SimulationRunLoadBalancerEnabled && status == SimulationRunStatus.ready)
+            {
+                var readyRuns = listResult.Items.ToList();
```
polomani, line 150: move this line out as well, code breaks

Cognite.Simulator.Utils/SimulationRunnerBase.cs
```diff
@@ -153,11 +139,34 @@ private async Task<IEnumerable<SimulationRun>> FindSimulationRunsWithStatus(
...
+                            new SimulationRunPollItem()
+                            {
+                                SimulatorIntegrationExternalId = connectorName,
+                                Limit = remainingLimit
+                            }
+                        ], token).ConfigureAwait(false);
+                    readyRuns.AddRange(pollResult.Items);
+                }
+
+                return readyRuns;
```
polomani, line 166: no need to have this return here, change the bottom return to return `readyRuns`

polomani (approved):
🦄

PR #337, opened 2026-05-11
feat: Add 'Parsing' status for model revision

Hmnt39:
When `SimulationRunLoadBalancingEnabled = true`, the connector now sets a model revision's status to parsing in CDF immediately before calling `ExtractModelInformation`. This acts as a distributed soft-lock: if another connector instance sees a revision already in parsing and it was updated within `ModelParsingTimeout` seconds, it skips it. If the revision has been stuck in parsing past that threshold (indicating a crashed connector), it re-parses and takes over.
When the flag is off, the old behavior is fully preserved — `ExtractModelInformation` is called directly with no status updates.

ModelLibraryConfig: added ModelParsingTimeout (default 3600s) — hard timeout for parsing; also used as the staleness threshold to detect and recover from crashed connectors stuck in parsing


Cognite.Simulator.Utils/ModelLibraryBase.cs
```diff
@@ -406,11 +406,44 @@ private async Task ExtractModelInformationAndPersist(T modelState, CancellationT
                     try
                     {
                         _logger.LogInformation("Extracting model information for {ModelExtid} v{Version}", modelState.ModelExternalId, modelState.Version);
-                        await ExtractModelInformation(modelState, token).ConfigureAwait(false);
+
+                        // Set the status to "Parsing" and persist it to CDF
```
mohammad-arif662: this comment and similar below not needed as they are not providing extra information

Cognite.Simulator.Utils/ModelLibraryBase.cs
```diff
@@ -406,11 +406,44 @@ private async Task ExtractModelInformationAndPersist(T modelState, CancellationT
...
+                        modelState.ParsingInfo.SetParsing();
+                        await PersistModelParsingStatus(modelState, token).ConfigureAwait(false);
```
mohammad-arif662: can we reuse existing method `PersistModelStatus` ? seems exactly same because in `PersistModelParsingStatus` method this condition `modelState.ParsingInfo.Status == SimulatorModelRevisionStatus.parsing` is always true as it is called right after `modelState.ParsingInfo.SetParsing();`

Cognite.Simulator.Utils/ModelLibraryBase.cs
```diff
@@ -431,6 +464,43 @@ await _cdfSimulatorResources.UpdateSimulatorModelRevisionParsingStatus(
...
+        /// Handles parsing timeout by setting the model status to "failure".
+        /// The failure status will be persisted to CDF by the caller's finally block via <see cref="PersistModelStatus"/>.
+        /// </summary>
+        /// <param name="modelState">Model file state</param>
+        private void HandleParsingTimeout(T modelState)
+        {
+            _logger.LogWarning("Model parsing exceeded timeout for {ModelExtid} v{Version}. Setting status to failure",
+                modelState.ModelExternalId, modelState.Version);
+
+            modelState.ParsingInfo.SetParsingTimeout();
```
mohammad-arif662: we can use `modelState.ParsingInfo.SetFailure("Model parsing timed out");` as both are setting to failure status with different error message. so `SetParsingTimeout` it will not be needed

Cognite.Simulator.Utils/ModelLibraryBase.cs
```diff
@@ -431,6 +464,43 @@ await _cdfSimulatorResources.UpdateSimulatorModelRevisionParsingStatus(
...
+        }
+
+        /// <summary>
+        /// Handles parsing timeout by setting the model status to "failure".
+        /// The failure status will be persisted to CDF by the caller's finally block via <see cref="PersistModelStatus"/>.
+        /// </summary>
+        /// <param name="modelState">Model file state</param>
+        private void HandleParsingTimeout(T modelState)
+        {
+            _logger.LogWarning("Model parsing exceeded timeout for {ModelExtid} v{Version}. Setting status to failure",
```
mohammad-arif662: message could be like this ? `Model parsing timed out for {ModelExtid} v{Version}",
                modelState.ModelExternalId, modelState.Version`

Cognite.Simulator.Utils/ModelLibraryBase.cs
```diff
@@ -406,11 +406,44 @@ private async Task ExtractModelInformationAndPersist(T modelState, CancellationT
                     try
                     {
                         _logger.LogInformation("Extracting model information for {ModelExtid} v{Version}", modelState.ModelExternalId, modelState.Version);
-                        await ExtractModelInformation(modelState, token).ConfigureAwait(false);
+
+                        // Set the status to "Parsing" and persist it to CDF
+                        modelState.ParsingInfo.SetParsing();
```
mohammad-arif662: New logic to set status to parsing should be behind this configuration flag `SimulationRunLoadBalancingEnabled`

if(SimulationRunLoadBalancingEnabled){
    // new parsing status logic
}else{
   // old logic
}

Cognite.Simulator.Utils/ModelLibraryBase.cs
Hmnt39: Removed the comments

Cognite.Simulator.Utils/ModelLibraryBase.cs
Hmnt39: Make sense

Cognite.Simulator.Utils/ModelLibraryBase.cs
Hmnt39: Chnages done

Cognite.Simulator.Utils/ModelLibraryBase.cs
Hmnt39: Changes done

Cognite.Simulator.Utils/ModelLibraryBase.cs
Hmnt39: Added the code inside flag check

Cognite.Simulator.Utils/ModelLibraryBase.cs
```diff
@@ -406,7 +429,37 @@ private async Task ExtractModelInformationAndPersist(T modelState, CancellationT
...
+                                    _logger.LogWarning("Model parsing timed out for {ModelExtid} v{Version}. Setting status to failure",
+                                        modelState.ModelExternalId, modelState.Version);
+                                    modelState.ParsingInfo.SetFailure("Model parsing timed out");
+                                    return;
+                                }
+                                catch (Exception e) when (e is not OperationCanceledException)
+                                {
+                                    _logger.LogError(e, "Model parsing failed for {ModelExtid} v{Version}",
+                                        modelState.ModelExternalId, modelState.Version);
+                                    modelState.ParsingInfo.SetFailure(e.Message);
```
vikramlc-cognite, line 454: Question: Why is that there is no return here while line 449 has a return?

Cognite.Simulator.Tests/UtilsTests/ModelLibraryMockedTest.cs
vikramlc-cognite, line 1: Can we add a test for the scenario read where revision last updated 2h ago, timeout is 1h and re-parse is initiated ? The current test does not cover any such scenario.

Cognite.Simulator.Utils/ModelLibraryBase.cs
mohammad-arif662: removed return so it calls `PersistModelInformation` in both cases

vikramlc-cognite (dismissed):

vikramlc-cognite (approved):

Cognite.Simulator.Tests/UtilsTests/TestUtilities.cs
```diff
@@ -399,13 +422,20 @@ public static void CleanUpFiles(ModelLibraryConfig? modelLibraryConfig, StateSto
...
+                Action<DefaultConfig<AutomationConfig>>? configModifier = null,
+                Func<DefaultModelFilestate, CancellationToken, Task>? extractModelInfoOverride = null)
         {
             var mockedLogger = new Mock<ILogger<DefaultModelLibrary<AutomationConfig, DefaultModelFilestate, DefaultModelFileStatePoco>>>();
             var mockedSimulatorClient = new Mock<ISimulatorClient<DefaultModelFilestate, SimulatorRoutineRevision>>();
             mockedSimulatorClient.Setup(client => client.ExtractModelInformation(It.IsAny<DefaultModelFilestate>(), It.IsAny<CancellationToken>()))
                 .Returns((DefaultModelFilestate state, CancellationToken token) =>
                 {
+                    if (extractModelInfoOverride != null)
+                        return extractModelInfoOverride(state, token);
```
polomani, line 438: passing a custom `mockedSimulatorClient` with "hanging" `ExtractModelInformation` method from `SetupRuntime` from your test would have been cleaner, I believe.

polomani: this is a bit cryptic to read!

Cognite.Simulator.Utils/ModelLibraryBase.cs
```diff
@@ -293,6 +297,25 @@ private async Task<T> GetOrAddModelRevisionImpl(string modelRevisionExternalId,
 
             if (downloaded && modelState.ShouldProcess())
             {
+                if (_connectorConfig.SimulationRunLoadBalancingEnabled
```
polomani, line 300: I don't believe this should be gated by simulation run load balancing flag, as it sounds like a separate feature. but this is up to the team, it's just weird reading the `connectorConfig` in model library

 @mohammad-arif662 why do you think it has to be tied to simulation runs balancer?
If you need to make it safer by disabling, you could add new feature flag on `ModelLibraryConfig` qith default=true, but for it seems like the feature that could be enabled for anyone installing the new version of the connector

polomani: anyway, you could keep it, it's just a bit surprising

Cognite.Simulator.Tests/UtilsTests/TestUtilities.cs
```diff
@@ -334,6 +335,28 @@ public static HttpResponseMessage MockSimulatorModelRevEndpoint()
             return OkItemsResponse(item);
         }
 
+        public static HttpResponseMessage MockSimulatorModelRevParsingEndpoint(long? lastUpdatedTime = null)
```
polomani, line 338: nit: you could extend `MockSimulatorModelRevEndpoint` instead by adding some optional params to it (status, lastUpdatedTime)

polomani (commented):
🦄 

I left some suggestions, but it's up to you whether you want to fix them

mohammad-arif662:
> 🦄
> 
> I left some suggestions, but it's up to you whether you want to fix them

sure, will raise a follow up PR to address above comments

PR #344, opened 2026-07-01
fix: clean up partial downloads and handle model revision failures in model library

mohammad-arif662:
- **Fix partial download leaving corrupt file on disk (`FileStorageClient.cs`)**: When a network interruption raised `IOException` during `DownloadFileAsync`, the partially written file was left on disk. On the next iteration, `File.Exists` returned `true` so the connector skipped re-downloading and attempted to extract a truncated ZIP, failing on every retry. The fix deletes the partial file in the `IOException` catch block so the next tick triggers a clean re-download.

PR in simconnect-dotnet repo to release new connector verion with this fix : https://github.com/cognitedata/simconnect-dotnet/pull/426

Jira: https://cognitedata.atlassian.net/browse/EDG-616


Cognite.Simulator.Utils/ModelLibraryBase.cs
```diff
@@ -634,12 +634,19 @@ private async Task FindModelRevisions(
...
+                        }
+                    }
+                    catch (Exception ex)
                     {
-                        _libState.UpdateDestinationRange(
-                            CogniteTime.FromUnixTimeMilliseconds(state.UpdatedTime),
-                            CogniteTime.FromUnixTimeMilliseconds(state.UpdatedTime));
+                        _logger.LogError("Failed to process model revision {ExternalId}: {Message}", revision.ExternalId, ex.Message);
                     }
                 }
```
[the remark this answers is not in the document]
mohammad-arif662: revision comes from modelRevisionsAllRes.Items which is deserialized from a CDF API response — null items don't occur in practice. Added exception object ex as the first argument to _logger.LogError

Cognite.Simulator.Utils/ModelLibraryBase.cs
```diff
@@ -634,12 +634,19 @@ private async Task FindModelRevisions(
 
                 foreach (var revision in modelRevisionsRes)
                 {
-                    var state = await GetOrAddModelRevision(revision.ExternalId, cachedRevisions, token).ConfigureAwait(false);
-                    if (state != null && state.Downloaded)
+                    try
+                    {
+                        var state = await GetOrAddModelRevision(revision.ExternalId, cachedRevisions, token).ConfigureAwait(false);
```
vikramlc-cognite: - The new catch (Exception ex) also catches the **please stop now** signal (cancellation), not just real errors.
- So when the connector is told to shut down, instead of stopping immediately, it logs an error for every remaining revision and keeps looping before it finally exits.

```
// Current (catches everything, including cancellation)
catch (Exception ex)
{
    _logger.LogError(ex, "Failed to process model revision {ExternalId}", revision.ExternalId);
}

// Better (let cancellation pass through, only swallow real failures)
catch (Exception ex) when (ex is not OperationCanceledException)
{
    _logger.LogError(ex, "Failed to process model revision {ExternalId}", revision.ExternalId);
}
```

Cognite.Simulator.Utils/ModelLibraryBase.cs
vikramlc-cognite, line 1: No test proves the **one bad revision doesn't block the rest** fix actually works. Kindly add the test for the same.

CHANGELOG.md
```diff
@@ -2,6 +2,12 @@
 
 All notable changes to this project will be documented in this file.
 
+## Release v1.0.0-beta-031 (2026-07-01)
+
+### Bug fixes
+
+- Delete partially written file when an `IOException` occurs during download, preventing corrupt ZIPs from being retried indefinitely.
+
```
vikramlc-cognite, line 10: Also add below:
`Isolate model revision processing failures so one bad revision no longer blocks subsequent revisions from being processed.`

Cognite.Simulator.Utils/ModelLibraryBase.cs
mohammad-arif662: added test

vikramlc-cognite (dismissed):

Cognite.Simulator.Utils/ModelLibraryBase.cs
```diff
@@ -634,12 +634,19 @@ private async Task FindModelRevisions(
...
+                                CogniteTime.FromUnixTimeMilliseconds(state.UpdatedTime),
+                                CogniteTime.FromUnixTimeMilliseconds(state.UpdatedTime));
+                        }
+                    }
+                    catch (Exception ex) when (ex is not OperationCanceledException)
                     {
-                        _libState.UpdateDestinationRange(
-                            CogniteTime.FromUnixTimeMilliseconds(state.UpdatedTime),
-                            CogniteTime.FromUnixTimeMilliseconds(state.UpdatedTime));
+                        _logger.LogError(ex, "Failed to process model revision {ExternalId}", revision.ExternalId);
```
polomani: I won't recommend catching this here to "let other model revisions to succeed", as this means if there was an intermittent error (like IOException) and the next revision succeeds you will never retry this model revision until there is a connector restart. 

Due to this filter 
```
Filter = new SimulatorModelRevisionFilter()
  {
      SimulatorExternalIds = [_simulatorDefinition.ExternalId],
      LastUpdatedTime = new CogniteSdk.TimeRange() { Min = updatedAfter + 1 },
  }
```

I advice reverting this try catch unless you want to get into the rabbit hole of fixing a lot of other regressions due to this change.

CHANGELOG.md
```diff
@@ -2,6 +2,13 @@
 
 All notable changes to this project will be documented in this file.
 
+## Release v1.0.0-beta-031 (2026-07-01)
+
+### Bug fixes
+
+- Delete partially written file when an `IOException` occurs during download, preventing corrupt ZIPs from being retried indefinitely.
+- Isolate model revision processing failures so one bad revision does not block subsequent revisions from being processed.
```
polomani: don't do this. intermittent errors need to be retried

Cognite.Simulator.Utils/FileStorageClient.cs
```diff
@@ -90,6 +90,17 @@ public async Task<bool> DownloadFileAsync(Uri uri, string filePath)
             catch (IOException e)
             {
                 _logger.LogError("I/O error occurred while saving the file into {filePath}: {Message}", filePath, e.Message);
+                try
+                {
+                    if (File.Exists(filePath))
+                    {
+                        File.Delete(filePath);
+                    }
```
polomani: you have these lines above as well, move them into a function

polomani: @mohammad-arif662 almost every PR I remind you about copy-paste..

Cognite.Simulator.Tests/UtilsTests/FileStorageClientUnitTest.cs
```diff
@@ -24,6 +27,61 @@ HttpResponseMessage MockFilesDownloadEndpointMaxSize()
...
+            content.Headers.ContentLength = 100;
+            return new HttpResponseMessage(HttpStatusCode.OK) { Content = content };
+        }
+
+        private sealed class ThrowingOnReadStream : Stream
+        {
+            public override bool CanRead => true;
+            public override bool CanSeek => false;
+            public override bool CanWrite => false;
+            public override long Length => throw new NotSupportedException();
```
polomani, line 74: may you could override FileStream instead so you get most members for free?

```
private sealed class ThrowingOnReadStream : FileStream
        {
            public ThrowingOnReadStream() : base(Path.GetTempFileName(), FileMode.Open, FileAccess.Read) { }
            public override Task<int> ReadAsync(byte[] buffer, int offset, int count, CancellationToken cancellationToken)
                => Task.FromException<int>(new IOException("Simulated I/O error during stream read"));
```

Cognite.Simulator.Utils/FileStorageClient.cs
polomani: StateUtils.DeleteLocalFile

Cognite.Simulator.Tests/UtilsTests/FileStorageClientUnitTest.cs
mohammad-arif662: `FileStream` needs a real file on disk and `Path.GetTempFileName()` will create a file every time, we can override `Dispose` method to clean it up but it will make same amount of code again.

Cognite.Simulator.Utils/ModelLibraryBase.cs
mohammad-arif662: This is added to fix a issue in this [PR](https://github.com/cognitedata/simconnect-dotnet/pull/426/changes#diff-998fd6a5ac4c0379fb2f616a2ec6539d64903e0317ddd65ef663d3f3de674306R244) added a throw in `ExtractModelInformation`. Issue is when GAP simulation run triggers revision extraction and extraction fails because of which `ProductionFilePath` is null and when it runs `GAP.OPENFILE(null)` and crashes with a misleading `ArgumentException: File path cannot be null or empty` hiding the actual error, that is because extraction failed.

mohammad-arif662: Even without this try catch failed revision updates `LastUpdatedTime` with this code

`_libState.UpdateDestinationRange(
                                CogniteTime.FromUnixTimeMilliseconds(state.UpdatedTime),
                                CogniteTime.FromUnixTimeMilliseconds(state.UpdatedTime));`
                                
                                
                                
 

   and will not be fetched in next loop to filter revisions to retry.

Cognite.Simulator.Tests/UtilsTests/FileStorageClientUnitTest.cs
polomani: ok!

Cognite.Simulator.Utils/ModelLibraryBase.cs
polomani: > This is added to fix a issue in this [PR](https://github.com/cognitedata/simconnect-dotnet/pull/426/changes#diff-998fd6a5ac4c0379fb2f616a2ec6539d64903e0317ddd65ef663d3f3de674306R244) added a throw in ExtractModelInformation

you are changing simutils to support one particular code path in GAP, this is not how this was designed initially, `ExtractModelInformation` is not supposed to throw errors, now you basically introduce the whole chain of changes because the behaviour has changed, the next one will be PROSPER, Symmetry, PetroSIM and so on, I don't think you'd like to go into that rabbit hole.

If the error is not propagating when `ProductionFilePath is null` you should throw error inside the callstack of Gap and not change the pattern for everyone. E.g. overriding `GetModelFilePath` or similar as you did (but also catch it in GAP before it escapes ExtractModelInformation)

polomani: In short, you are changing to many things at the same time, this can lead to way more edge cases.

polomani: cleaning up the file on error while downloading is a good change, everything else is risky

mohammad-arif662: yes, I will revert it because extraction failed because file was corrupted and now with this change of deleting file not fully downloaded it will not be corrupted. And also added change [GetModelFilePath](https://github.com/cognitedata/simconnect-dotnet/pull/426/changes#diff-4f6ae8f8527eb7e5de63c02200a6cb608e4e214c25873659775fdebc8a547b48L247)  to fail simulation when `ProductionFilePath` in null.

polomani:
🦄

vikramlc-cognite (approved):

